PR #4, opened 2020-02-27
fixed error when retrieving tasks (#2)

### LINKED ISSUE #2: Error when retrieving tasks for a namespace
When expanding a namespace node and the list of tasks is retrieved this error is printed 

java.io.IOException: Process exited with an error: 1 (Exit value: 1) Error: unable to match a printer suitable for the output format "'jsonpath={range .items[*]}{.metadata.name}{\'\\n\'}", allowed formats are: go-template,go-template-file,json,jsonpath,jsonpath-file,name,template,templatefile,yaml

lstocchi:
it fixes #2

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/tree/TasksNode.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/tree/TasksNode.java
@@ -26,7 +26,7 @@ public class TasksNode extends LazyMutableTreeNode implements IconTreeNode {
         super.load();
         try {
             NamespaceNode namespaceNode = (NamespaceNode) getParent();
-            ((TektonRootNode)getRoot()).getTkn().getTasks(namespaceNode.toString()).forEach(task -> add(new TaskNode(task)));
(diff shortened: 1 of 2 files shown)


src/main/java/com/redhat/devtools/intellij/tektoncd/tree/TasksNode.java
```diff
@@ -26,7 +26,7 @@ public void load() {
         super.load();
         try {
             NamespaceNode namespaceNode = (NamespaceNode) getParent();
-            ((TektonRootNode)getRoot()).getTkn().getTasks(namespaceNode.toString()).forEach(task -> add(new TaskNode(task)));
+            ((TektonRootNode)getRoot()).getTkn().getTasks(namespaceNode.toString()).forEach(task -> { if(!task.isEmpty()) add(new TaskNode(task)); });
```
jeffmaury: I think this should be done at the CLI level, the result of getTasks should not return empty names !!!. Using Java Streams could filter the list

PR #5, opened 2020-02-27
Added support for pipelines/resources/clustertasks nodes

### LINKED ISSUE #3: Add support for pipeline/pipelineResources/clusterTask nodes
The tree only supports Namespace and Task nodes. We need to add those other items and list them on tree. Not sure if we also need to create nodes for volume/workspaces (maybe in future??)

lstocchi:
It fixes #3 
Now the tree also shows pipelines, pipeline resources and clustertasks. We need to find a new icon for resources (copied how it is now in vscode).

image


src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/PipelineRun.java
```diff
@@ -0,0 +1,33 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Red Hat, Inc.
```
jeffmaury: Should be 2020

src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/PipelineRunDeserializer.java
```diff
@@ -0,0 +1,42 @@
+package com.redhat.devtools.intellij.tektoncd.tkn;
```
jeffmaury: Add header

src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCli.java
```diff
@@ -79,6 +78,26 @@ public boolean isTektonAware(KubernetesClient client) {
...
+    public List<String> getPipelines(String namespace) throws IOException {
+        String output = ExecHelper.execute(command, "pipeline", "ls", "-n", namespace, "-o", "jsonpath={.items[*].metadata.name}");
+        return Arrays.stream(output.split("\\s+")).filter(item -> !item.isEmpty()).collect(Collectors.toList());
+    }
+
+    @Override
+    public List<PipelineRun> getPipelineRuns(String namespace, String pipeline) throws IOException {
+        String json = ExecHelper.execute(command, "pipelinerun", "ls", pipeline, "-n", namespace, "-o", "json");
+        return JSON_MAPPER.
+                registerModule(new SimpleModule().addDeserializer(List.class, new PipelineRunDeserializer())).
```
jeffmaury: I'm wondering if registering a new module for each call does not lead to memory leaks

src/main/java/com/redhat/devtools/intellij/tektoncd/tree/ClusterTasksNode.java
```diff
@@ -0,0 +1,29 @@
+package com.redhat.devtools.intellij.tektoncd.tree;
```
jeffmaury: Add header

src/main/java/com/redhat/devtools/intellij/tektoncd/tree/PipelineNode.java
```diff
@@ -0,0 +1,40 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Red Hat, Inc.
```
jeffmaury: Should be 2020

src/main/java/com/redhat/devtools/intellij/tektoncd/tree/PipelineRunNode.java
```diff
@@ -0,0 +1,32 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Red Hat, Inc.
```
jeffmaury: Should be 2020

src/main/java/com/redhat/devtools/intellij/tektoncd/tree/PipelinesNode.java
```diff
@@ -0,0 +1,39 @@
+package com.redhat.devtools.intellij.tektoncd.tree;
+
+/*******************************************************************************
```
jeffmaury: Header is first

src/main/java/com/redhat/devtools/intellij/tektoncd/tree/PipelinesNode.java
```diff
@@ -0,0 +1,39 @@
...
+ * Copyright (c) 2019 Red Hat, Inc.
```
jeffmaury: Should be 2020

src/main/java/com/redhat/devtools/intellij/tektoncd/tree/ResourceNode.java
```diff
@@ -0,0 +1,15 @@
+package com.redhat.devtools.intellij.tektoncd.tree;
```
jeffmaury: Add header

src/main/java/com/redhat/devtools/intellij/tektoncd/tree/ResourcesNode.java
```diff
@@ -0,0 +1,29 @@
+package com.redhat.devtools.intellij.tektoncd.tree;
```
jeffmaury: Add header

src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCli.java
lstocchi: I saw you already created a PR on fabric8 repo to solve the issue you found and they seem to answer fastly. For the moment i came back to the static block so we should be safe with any possible memory leaks. When the issue on fabric8 will be solved we can move to that one

src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCli.java
```diff
@@ -19,20 +19,23 @@
...
-        SimpleModule module = new SimpleModule();
-        module.addDeserializer(List.class, new TaskRunDeserializer());
-        JSON_MAPPER.registerModule(module);
+        SimpleModule tr_module = new SimpleModule();
+        tr_module.addDeserializer(List.class, new TaskRunDeserializer());
+        TASKRUN_JSON_MAPPER.registerModule(tr_module);
+
+        SimpleModule pr_module = new SimpleModule();
+        pr_module.addDeserializer(List.class, new PipelineRunDeserializer());
+        PIPERUN_JSON_MAPPER.registerModule(pr_module);
```
jeffmaury: Note sure you need 2 distincts mappers, only one with 2 registers serializers should be ok but approving as I'm not sure

lstocchi: By using a single mapper it always uses the last deserializer added so when expanding a node you can end up having an issue like "pipelinerun node cannot be cast to taskrun..."

PR #13, opened 2020-02-28
added run task action (#11)

### LINKED ISSUE #11: Run a task
Users should be able to run a task by using the extension

lstocchi:
it fixes #11 

2 thoughts here. First, this action just execute the start task cmd, do we need to print something somewhere? Second, the tree doesn't get refreshed automatically, do we need to add support for watches so it can listen to updates from cluster (creation/update of new taskrun)? Or for a first release a refresh btn is enough?


src/main/java/com/redhat/devtools/intellij/common/actions/component/RunTaskAction.java
```diff
@@ -0,0 +1,29 @@
...
+
+import javax.swing.tree.TreePath;
+import java.io.IOException;
+
+public class RunTaskAction extends TknAction {
+    public RunTaskAction() { super(TaskNode.class); }
+
+    @Override
+    public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, String tkncli) throws IOException {
+        ExecHelper.execute(tkncli, "task", "start", selected.toString(), "-n", ((TaskNode)selected).getParent().getParent().toString());
```
jeffmaury: Every run of the tkn cli should be done in the TknCli class and not spread all over the code. Also as a new task run will be created, you must class refresh on the selected object so that the list will be refreshed. And the command string should be hidden and Tkn object should be given to any action

src/main/java/com/redhat/devtools/intellij/common/actions/TknAction.java
```diff
@@ -0,0 +1,33 @@
...
+    @Override
+    public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected) {
+        try {
+            this.actionPerformed(anActionEvent, path, selected, ((TknCli) TknCli.get()).getCommand());
+        } catch (IOException e) {
+            Messages.showErrorDialog("Error: " + e.getLocalizedMessage(), "Error");
+        }
+    }
+
+    public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, String tkncli) throws IOException { }
```
jeffmaury: Should be Tkn tkncli insted of String tkncli

src/main/java/com/redhat/devtools/intellij/common/actions/component/RunTaskAction.java
lstocchi: Reloading the selected object doesn't solve the issue of having an updated list. The last launched taskrun doesn't appear even if we reload the node (creating a pod and run the task requires some time) and, in any case, the user will see the task in a running state and the tree won't refresh when the task succeed or fail. So the resulting list will always be not updated

lstocchi:
@siamaksade we are implementing the `run task` feature and we'd like to have your feedback on this. Our idea was to show a form where the user could set inputs/outputs before running a task. 
Below it is a sample of what we have now (not yet finished but it can give you an idea of the final work). There are 3 different sections: inputs, outputs and preview. 

- Inputs contains a combobox which displays all inputs the task requires to run (params and resources). Based on its kind, different textboxes/combobox are shown (in the example below `pathToDockerFile` is a param and `workspace` a resource).
- Output contains a combobox which displays all outputs specified in the task
- Preview allows users to check the value already inserted through the form and see all inputs/outputs updated in JSON or YAML form (in the gif there are no outputs, still to be implemented).

I also wanted to add a button which opens the TaskRun configuration in an editor so user can also edit the configuration by hand.

The other solution would be to create a table and show all inputs/outputs in different rows.

WDYT? Any idea how we could improve the workflow? I'm also available to show you real-time in a demo. Just let me know.

ij-tekton-taskrun

lstocchi:
Moved to YAML
image

What should we do when no outputs have to be set? This is how it's right now
image

jeffmaury:
The no inputs looks good to me. However, what is the refresh button on the right. I think the YAML view must be automatically refreshed when the user updates inputs or outputs.

lstocchi:
Yeah YAML view is updated automatically and also refreshed but having a refresh button can help people with this. Currently I update the YAML when the textbox (in the case above the value textbox) lose the focus. In this way it means that the user finished adding everything he was typing and went to set another input/output or run the task.

The lose focus is triggered if the user clicks on another part of the form which is not the textbox (he clicks on the combo to change input, he clicks on the preview area....) but having a refresh button for me it's more clear and natural. If the user is lost and doesn't know that he needs to click outside the textbox, he can always click on the refresh button. 

So even if it's not mandatory to click the refresh btn i would keep it to have the ui more user friendly.

lstocchi:
@jeffmaury The dialog is done. The logic is all there. Let's wait what Siamak think about this.

The only problem for me is that there is no way to setting a different path for a resource and start the task using the cli so the dialog is showing a useless textbox atm. Next sprint we need to clarify this with the tekton team.

siamaksade:
@lstocchi although the dialog is correctly reflecting the TaskRun spec, it might be more intuitive to separate params from inputs. The dialog would have 3 sections then:
- **Parameters**: list of params and a textbox for the value of each param
- **Input Resources**: list of input resources and a dropbox for each resource, listing available pipelineresources matching the type of the input resource
- **Output Resources**: list of output resources and a dropbox for each resource, listing available pipelineresources matching the type of the output resource

Regarding overriding resource paths, I'd remove it from the dialog for now and create an issue for adding it to the CLI: https://github.com/tektoncd/cli

It might also be worth to add support for workspaces:
https://github.com/tektoncd/pipeline/blob/master/docs/taskruns.md#workspaces

It's not available in VS Code and Console yet but it will be added there soon.

lstocchi:
@siamaksade ok. I'm going to split params with input resources and remove the resource paths txtbox for now. I'll open an issue here and in the cli repo so we can work on it later on.

Regarding workspaces i am going to open another issue here so we can add support for them as soon as we release the first version of this extension.

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/StartTaskDialogAction.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.concurrent.CompletableFuture;
+
+public class StartTaskDialogAction extends TektonAction {
+    Logger logger = LoggerFactory.getLogger(StartTaskDialogAction.class);
+
+    public StartTaskDialogAction() { super(TaskNode.class); }
+
+    @Override
+    public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, Tkn tkncli) {
+        String namespace = ((TaskNode)selected).getParent().getParent().toString();
```
jeffmaury: By definition any action is running in the UI thread so there is no need to the executeInUI. But the bad part in that when you create the dialog, you will call tkncli that may block the UI. So I would rewrite as:
ExecHelper.submit(() -> {
call tkncli to get info the the dialog
UIHelper.executeInUI() -> create dialog and retrieve result)
execute the run task (as we are in a separate thread)
}

src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/Tkn.java
```diff
@@ -142,4 +142,14 @@
...
 
+    /**
+     * Start the execution of a task
+     *
+     * @param namespace the namespace of the task
+     * @param task the task that has to be run
+     * @param args args to start task (contains input/output params/resources)
+     * @throws IOException if communication errored
+     */
+    void runTask(String namespace, String task, String args) throws IOException;
```
jeffmaury: I don't like args being passed as a single parameter. You will face issues when parameters have space in there values. Should be an an array or a collection (list/set)

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/StartTaskDialogAction.java
```diff
@@ -0,0 +1,61 @@
...
+import com.redhat.devtools.intellij.tektoncd.tree.TaskNode;
+import com.redhat.devtools.intellij.tektoncd.ui.component.StartTaskDialog;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.swing.tree.TreePath;
+import java.io.IOException;
+import java.util.List;
+
+public class StartTaskDialogAction extends TektonAction {
```
jeffmaury: To be unifom, should be renamed as StartTaskAction

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/StartTaskDialogAction.java
```diff
@@ -0,0 +1,61 @@
...
+            try {
+                task = tkncli.getTaskYAML(namespace, selected.toString());
+                resources = tkncli.getResources(namespace);
+            } catch (IOException e) {
+                logger.error("Error: " + e.getLocalizedMessage());
+                return;
+            }
+            StartTaskDialog stdialog = UIHelper.executeInUI(() -> {
+                StartTaskDialog dialog = new StartTaskDialog(null, task, resources);
+                dialog.show();
```
jeffmaury: If the user cancel the dialog, we should stop the action and do nothing

src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/Tkn.java
```diff
@@ -166,4 +166,15 @@
...
+
+    /**
+     * Start the execution of a task
+     *
+     * @param namespace the namespace of the task
+     * @param task the task that has to be run
+     * @param args args to start task (contains input/output params/resources)
+     * @throws IOException if communication errored
+     */
+    void startTask(String namespace, String task, List<String> args) throws IOException;
```
jeffmaury: This array of strings is too much linked to the CLI. I would rather prefer having an interface like:
`startTask(String namespace, String task, Map<String, String> parameters, Map<String, String> inputResources, Map<String, String> outputResources)`.
You are putting some CLI logic into the dialog (-i, -p or -o)

src/main/java/com/redhat/devtools/intellij/tektoncd/ui/component/StartTaskDialog.java
```diff
@@ -0,0 +1,459 @@
...
+
+    public List<String> args() {
+        return args;
+    }
+
+    private void calculateArgs() {
+        if (inputs != null) {
+            for (Input input : inputs) {
+                if (input.kind() == Input.Kind.PARAMETER) {
+                    String value = input.value() == null ? input.defaultValue().get() : input.value();
```
jeffmaury: defaultValue is an Optional so we may get an exception here

src/main/java/com/redhat/devtools/intellij/tektoncd/tree/ResourcesNode.java
```diff
@@ -26,7 +26,7 @@ public void load() {
         super.load();
         try {
             NamespaceNode namespaceNode = (NamespaceNode) getParent();
-            ((TektonRootNode)getRoot()).getTkn().getResources(namespaceNode.toString()).forEach(resource -> add(new ResourceNode(resource)));
+            ((TektonRootNode)getRoot()).getTkn().getResources(namespaceNode.toString()).forEach(resource -> add(new ResourceNode(resource.name())));
```
jeffmaury: I don't know the rationale for only keeping the name of the resource but I think it would be good to store the resource as we may want in the future to display other attributes (ex an icon based on the type of resource)

src/main/java/com/redhat/devtools/intellij/tektoncd/ui/component/StartTaskDialog.java
```diff
@@ -0,0 +1,459 @@
...
+                // when outputsResourcesCB combo box value changes, the new value is saved and preview is updated
+                int outputSelectedIndex = outputsCB.getSelectedIndex();
+                Resource resourceSelected = (Resource) itemEvent.getItem();
+                outputs.get(outputSelectedIndex).setValue(resourceSelected.name());
+                updatePreview();
+            }
+        });
+
+        // listener for when the button startTask is clicked
+        startTaskButton.addMouseListener(new MouseAdapter() {
```
jeffmaury: In order to be UI aligned with IntelliJ, your wrapper does not need to deal with the buttons and only create a center pane. As you need to rename the OK button from OK to Start or Start task, you can retrieve the OkAction and change the name attribute. For an example of such a dialog see: https://github.com/redhat-developer/intellij-openshift-connector/blob/master/src/main/java/org/jboss/tools/intellij/openshift/ui/service/CreateServiceDialog.java

src/main/java/com/redhat/devtools/intellij/tektoncd/ui/component/StartTaskDialog.java
```diff
@@ -0,0 +1,459 @@
...
+    private List<Output> outputs;
+
+    private String namespace;
+    private String taskName;
+    private List<String> args;
+
+    public StartTaskDialog(Component parent, String task, List<Resource> resources) {
+        super(null, parent, false, IdeModalityType.IDE);
+        try {
+            this.namespace = YAMLHelper.getStringValueFromYAML(task, new String[] {"metadata", "namespace"});
```
jeffmaury: Separation of concerns. The dialog implementation don't have to deal with YAML or JSON so it should be given an object representing the task. Maybe when we will switch to the Fabric8 Tekton API that would be more natural

src/main/resources/META-INF/plugin.xml
```diff
@@ -28,6 +28,7 @@
   <actions>
     <group id="com.redhat.devtools.intellij.tektoncd.tree" popup="true">
       <action class="com.redhat.devtools.intellij.tektoncd.actions.component.OpenEditorTaskAction" text="Open in Editor"/>
+        <action class="com.redhat.devtools.intellij.tektoncd.actions.component.StartTaskDialogAction" text="Start Task"/>
```
jeffmaury: As the action is on the Task node then text should be only Start (VSCode Tekton uses Start)

src/main/java/com/redhat/devtools/intellij/tektoncd/ui/component/StartTaskDialog.java
```diff
@@ -0,0 +1,459 @@
+/*******************************************************************************
```
jeffmaury: I suppose the `component` sub package comes from the OpenShift plugin but I think this should be renamed as `task` ?

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/StartTaskDialogAction.java
lstocchi: My idea was to create a new action and call it StartTaskAction. Right now the user can only use the dialog to set up the inputs/outputs but what if someone prefer to write everything in the editor? So i had in mind to add a new button which allows people to open the TaskRun on the editor and let them edit it. When it's finished they could "right-click" the file and select run task. Another point is if we follow what's happening in vscode. They are working to add two more nodes to display taskRun and pipelineRun separately. In that case we could use the run task action to just start the task with the inputs/outputs found in the taskrun. WDYT?

src/main/java/com/redhat/devtools/intellij/tektoncd/ui/component/StartTaskDialog.java
lstocchi: I guess you get the wrong example but i should have found the way to rename the OK btn. setOKButtonText and doOKAction. Are you referring to them?

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/StartTaskDialogAction.java
jeffmaury: Regarding the editor, you are messing 2 concerns: starting a task and updating it. Look at Jenkins there are 2 differents actions for a job (pipeline) start a build or edit the pipeline. And even if there are extra node, the concern is always to start a task to I think we should remove the Dialog from the class name to be inline with other IJ plugins

src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCli.java
```diff
@@ -170,4 +176,35 @@ public void editCustomResource(KubernetesClient client, String namespace, String
...
     }
+
+    public void startTask(String namespace, String task, Map<String, String> parameters, Map<String, String> inputResources, Map<String, String> outputResources) throws IOException {
+        List<String> args = new ArrayList<>(Arrays.asList("task", "start", task, "-n", namespace));
+        args.addAll(argsToList(parameters, inputResources, outputResources));
+        ExecHelper.execute(command, args.toArray(new String[0]));
+    }
+
+    private List<String> argsToList(Map<String, String> parameters, Map<String, String> inputResources, Map<String, String> outputResources) {
+        List<String> args = new ArrayList<>();
```
jeffmaury: There is probably some refactoring to be done here as each block is the same with the exception of the prefix so those blocks can be refactored as a separate method with prefix parameter and the map

Issue #10, opened 2020-02-28
Open pipeline in an editor


lstocchi:
Users should be allowed to check the pipeline configuration inside the ide

lstocchi:
This can be closed. PR was merged https://github.com/redhat-developer/intellij-tekton/pull/14

PR #14, opened 2020-03-02
add editor support for pipelines, tasks and resources (#10)

lstocchi:
it is related to #10 

This PR allows to open a pipeline/task/pipeline resource in the editor and see its configuration


lstocchi:
Changed a little bit how it works. By working on the updating/saving process i found out that working with LightVirtualFile doesn't trigger the onSave listener. Now a temporary file is created everytime a pipeline/task/resource is opened in the editor. The file is then deleted when the editor is closed. I also solved another problem bc before if you clicked on the same pipeline over and over again, it keeps opening several tabs with the same content. Now only one editor per component is opened.

src/main/java/com/redhat/devtools/intellij/common/actions/TknAction.java
```diff
@@ -0,0 +1,34 @@
+/*******************************************************************************
```
jeffmaury: The common package is supposed to be extracted from this plugin and should contain only generic stuff. So every piece of code that is Tekton specific must be moved under the tektoncd sub packages

src/main/java/com/redhat/devtools/intellij/common/actions/component/OpenEditorTask.java
```diff
@@ -0,0 +1,63 @@
...
+import com.redhat.devtools.intellij.tektoncd.tkn.Tkn;
+import com.redhat.devtools.intellij.tektoncd.tree.PipelineNode;
+import com.redhat.devtools.intellij.tektoncd.tree.ResourceNode;
+import com.redhat.devtools.intellij.tektoncd.tree.TaskNode;
+
+import javax.swing.tree.TreePath;
+import java.io.IOException;
+import java.util.Arrays;
+
+public class OpenEditorTask extends TknAction {
```
jeffmaury: Any reason why it does not followng the *Action name pattern

src/main/java/com/redhat/devtools/intellij/common/listener/FileEditorListener.java
```diff
@@ -0,0 +1,33 @@
...
+        if (tknProperty != null && tknProperty.equals("file")) {
+            ApplicationManager.getApplication().runWriteAction(new Runnable() {
+                @Override
+                public void run() {
+                    try {
+                        if (Files.exists(Paths.get(file.getPath()))) {
+                            file.delete(this);
+                        }
+                    } catch (IOException e) {
+                        e.printStackTrace();
```
jeffmaury: Replace by SLF4J API

lstocchi:
Removed the listener and so the messageBus to deal with that. The deleteOnExit works fine. Great idea!! @jeffmaury :+1:

PR #18, opened 2020-03-03
Add integration stage

jeffmaury:


lstocchi:
LGTM .... only one thing, shouldn't we also stop openshift `oc cluster down` at the end of the openshift integration test?

jeffmaury:
> 
> 
> LGTM .... only one thing, shouldn't we also stop openshift `oc cluster down` at the end of the openshift integration test?

Not absolutely required at the VM will be shut down but in order to be aligned with the Kubernetes stage I will add it

PR #19, opened 2020-03-05
Allow to update configuration in editor (#16)

### LINKED ISSUE #16: Allow to update configuration in editor
When #14 will be merged users will be able to check configurations they want but they will not be able to edit them. We should add this feature.

lstocchi:
it fixes #16 and it is built over #14


lstocchi:
I was just thinking that we don't make any check about the content and maybe we should have a confirmation from the user before to upload the content updated to the cluster. It may happen that the user have multiple tabs opened with incomplete edited pipeline/task/resource and if saveall is clicked, everything will be pushed on cluster. I think we should add a popup to ask a confirmation. i.e"Do you want to push the changes to the cluster?". WDYT? @jeffmaury

src/main/java/com/redhat/devtools/intellij/common/listener/FileEditorListener.java
```diff
@@ -0,0 +1,47 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+import static com.redhat.devtools.intellij.common.CommonConstants.TEKTON_RS;
+
+public class FileEditorListener implements FileEditorManagerListener {
```
jeffmaury: Either we keep it Tekton independant and remove the TEKTON constants and rename it to DeleteOnCloseListener. Another better solution would be to create the file in temporaty folder and then set the deleteOnExit flag so we don't have to get some specific code to handle.

src/main/java/com/redhat/devtools/intellij/common/listener/FileEditorListener.java
```diff
@@ -0,0 +1,47 @@
...
+    Logger logger = LoggerFactory.getLogger(FileEditorListener.class);
+    @Override
+    public void fileClosed(@NotNull FileEditorManager source, @NotNull VirtualFile file) {
+        String tknProperty = file.getUserData(TEKTON_RS);
+        if (tknProperty != null && !tknProperty.isEmpty()) {
+            ApplicationManager.getApplication().runWriteAction(new Runnable() {
```
jeffmaury: Why do you run it as a Write action ? Either you can call the delete method or you run it from the Application thread pool so not to block the UI thread but write actions are using a single lock

src/main/java/com/redhat/devtools/intellij/common/listener/SaveInEditorListener.java
```diff
@@ -0,0 +1,48 @@
+package com.redhat.devtools.intellij.common.listener;
```
jeffmaury: It is Tekton specific so should be moved out of the common package

src/main/java/com/redhat/devtools/intellij/common/CommonConstants.java
```diff
@@ -10,6 +10,9 @@
  ******************************************************************************/
 package com.redhat.devtools.intellij.common;
 
+import com.intellij.openapi.util.Key;
+
 public class CommonConstants {
     public static final String HOME_FOLDER = System.getProperty("user.home");
+    public static final Key<String> KIND_PLURAL = Key.create("tekton.plural");
```
jeffmaury: I don't think that should be in common package as it seems to be Tekton specific

src/main/java/com/redhat/devtools/intellij/tektoncd/listener/SaveInEditorListener.java
```diff
@@ -0,0 +1,77 @@
...
+                throw new IOException("Tekton file has not a valid format. Spec field is not found.");
+            }
+        } catch (IOException e) {
+            UIHelper.executeInUI(() -> Messages.showErrorDialog("Error: " + e.getLocalizedMessage(), "Error"));
+            return false;
+        }
+
+        Notification notification;
+        try (final KubernetesClient client = new DefaultKubernetesClient()) {
+            CustomResourceDefinitionContext crdContext = CRDHelper.getCRDContext(vf.getUserData(KIND_PLURAL));
```
jeffmaury: We have an issue with versions are you are hard coding v1alpha1 where Tekton is already switching to v1alpha2 so I think we should extract the CRD context from the resource file.

src/main/java/com/redhat/devtools/intellij/tektoncd/listener/SaveInEditorListener.java
```diff
@@ -0,0 +1,77 @@
...
+            JsonNode customResource = JSONHelper.MapToJSON(client.customResource(crdContext).get(namespace, name));
+            ((ObjectNode) customResource).set("spec", spec);
+            client.customResource(crdContext).edit(namespace, name, customResource.toString());
+        } catch (IOException e) {
+            // give a visual notification to user if an error occurs during saving
+            notification = new Notification("SaveNotification", "Error", "An error occurred while saving " + vf.getUserData(KIND_PLURAL) + " " + name + "\n" + e.getLocalizedMessage(), NotificationType.ERROR);
```
jeffmaury: See my comments on notifications

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/CRDHelper.java
```diff
@@ -0,0 +1,15 @@
+package com.redhat.devtools.intellij.tektoncd.utils;
```
jeffmaury: Add header

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/JSONHelper.java
```diff
@@ -0,0 +1,18 @@
+package com.redhat.devtools.intellij.tektoncd.utils;
```
jeffmaury: Add header and should be in common package as it's not Tekton specific

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/YAMLHelper.java
```diff
@@ -0,0 +1,32 @@
+package com.redhat.devtools.intellij.tektoncd.utils;
```
jeffmaury: Idem

src/main/java/com/redhat/devtools/intellij/tektoncd/listener/SaveInEditorListener.java
```diff
@@ -0,0 +1,96 @@
...
+            if (spec == null) {
+                throw new IOException("Tekton file has not a valid format. Spec field is not found.");
+            }
+        } catch (IOException e) {
+            UIHelper.executeInUI(() -> Messages.showErrorDialog("Error: " + e.getLocalizedMessage(), "Error"));
+            return false;
+        }
+
+        Notification notification;
+        try (final KubernetesClient client = new DefaultKubernetesClient()) {
```
jeffmaury: The Kubernetes client should not be created, it is costly and may requires tuning (proxy,...). So you should retrieve it from the root node. Also it would be good from OOP that every interaction with the cluster be implemented through the Tkn interface.

src/main/java/com/redhat/devtools/intellij/tektoncd/listener/SaveInEditorListener.java
```diff
@@ -0,0 +1,96 @@
...
+            JsonNode customResource = JSONHelper.MapToJSON(client.customResource(crdContext).get(namespace, name));
+            ((ObjectNode) customResource).set("spec", spec);
+            client.customResource(crdContext).edit(namespace, name, customResource.toString());
+        } catch (IOException e) {
+            // give a visual notification to user if an error occurs during saving
+            notification = new Notification(NOTIFICATION_ID, "Error", "An error occurred while saving " + vf.getUserData(KIND_PLURAL) + " " + name + "\n" + e.getLocalizedMessage(), NotificationType.ERROR);
```
jeffmaury: Be coherent with line 92, so also call StringUtils.capitalize

lstocchi: Heck if you are picky :grin:

lstocchi:
What's the point of saving a temporary file that the user cannot open again if he closes the editor page and gets deleted when IJ closes? I mean it should be the opposite, the last `return true;` is no sense. If we are going to save the file we are giving hope to the user that he can copy or retrieve the content in a second moment which is not true. He must be aware that if there is an error with the content of the tekton configuration or with the cluster connection he has to take action now.

src/main/java/com/redhat/devtools/intellij/tektoncd/listener/SaveInEditorListener.java
```diff
@@ -0,0 +1,138 @@
...
+        } catch (IOException e) {
+            notification = new Notification(NOTIFICATION_ID, "Error", "An error occurred while saving \n" + e.getLocalizedMessage(), NotificationType.ERROR);
+            Notifications.Bus.notify(notification);
+            logger.error("Error: " + e.getLocalizedMessage());
+            return false;
+        }
+
+        int resultDialog = UIHelper.executeInUI(() ->
+                Messages.showYesNoDialog(
+                        "The file has been saved. Do you want to push the changes to the cluster?",
```
jeffmaury: Shouldn't we change this label as the file may not be saved depending on the user response ?

lstocchi: Yes. What do you suggest?

lstocchi: Having "Do you want to push the changes to the cluster?" is enough?

lstocchi: Changed it so i can merge and rebase other PRs. If you have a better idea we can change it in the other PR related to save action

PR #21, opened 2020-03-09
Added delete action for pipelines/tasks/resources

### LINKED ISSUE #15: Delete Action
Users should be able to delete a component (pipeline/task/resource)

lstocchi:
it fixes #15


src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/DeleteAction.java
```diff
@@ -0,0 +1,66 @@
...
+                    String kind = selected.getClass().getSimpleName().toLowerCase().replace("node", "");
+                    DeleteDialog dialog = new DeleteDialog(null,
+                            "Delete " + selected.toString(),
+                            "Are you sure you want to delete " + kind + " " + selected.toString() + " ?");
+                    dialog.show();
+                    return dialog;
+                });
+                if (deleteDialog.isOK()) {
+                    String namespace = ((LazyMutableTreeNode)selected).getParent().getParent().toString();
+                    switch (selected.getClass().getSimpleName()) {
```
jeffmaury: Matching on SimpleName is not strict enough so you should use instead select.getClass().equals(PipelineNode.class) to test

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/DeleteAction.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.concurrent.CompletableFuture;
+
+public class DeleteAction extends TektonAction {
+    public DeleteAction() { super(TaskNode.class, PipelineNode.class, ResourceNode.class); }
+
+    @Override
+    public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, Tkn tkncli) throws IOException {
+        CompletableFuture.runAsync(() -> {
+            try {
+                DeleteDialog deleteDialog = UIHelper.executeInUI(() -> {
```
jeffmaury: You could move the dialog part out of the async block and run it in the UI thread and run only the cli part as an async block in order not to block the UI

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/TektonAction.java
```diff
@@ -39,6 +39,6 @@ private Tkn getTkn(AnActionEvent anActionEvent) throws IOException {
         return ((TektonRootNode)tree.getModel().getRoot()).getTkn();
     }
 
-    public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, Tkn tkn) {
+    public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, Tkn tkn) throws IOException {
```
jeffmaury: Not sure to understand why do you need the actions to throw IOException ?

jeffmaury:
Can you rebase it ?

Issue #17, opened 2020-03-02
Allow to create a pipeline or a task


lstocchi:
Looking at the VSCode extension i didn't find any action related to this but i think it could be useful to have a `create` btn which opens up an editor with a pipeline/task template to be completed.

lstocchi:
Closing it. The plugin now allows to create new pipelines (https://github.com/redhat-developer/intellij-tekton/pull/25), tasks/clustertasks/resources (https://github.com/redhat-developer/intellij-tekton/pull/35)

PR #25, opened 2020-03-17
Add support to create new pipeline and intellisense (#17)

lstocchi:
it is related to #17 

Here we have a big problem with the autosave. There is no way to find out if the saving was triggered by the user with the CTRL+S, or by clicking the `save all` in the menu or by the autosave. The only distinction IJ does is if the saving was made by the compilation (implicit) or by the autosave/user (explicit).
So if the autosave is not disabled the user can end up having multiple errors from the cluster bc, based on its settings, this can also be triggered every few seconds.

What i did was to create an action "PushToCluster" which gets triggered when the user press CTRL+S. So the action won't get triggered by the auto-save or if the user press the `save all` from the menu. After the action is triggered the user will be shown a popup where he will be asked if he wants to push the content of its file to the cluster.

The resource will then be created or updated.

The problem now is the tree that doesn't get refreshed. I don't have any access point to the tree from this action. Any idea? My first is to release a refresh action (as in VScode), second is to implement watch feature and leave to them all the job to refresh the tree, which is a problem we also have when running a task so we should solve 2 (or more) issues with one PR.

This PR also comes with basic intellisense stuff in it. It seemed to me quite useful to add it here. If i need to create a pipeline from scratch having some snippet it's helpful. Snippets are shown when either the title or any of the keyword listed in the file (copied from VSCode, i just edited it a bit) is typed in the editor.


lstocchi:
Edit: tree refresh added. After some reverse eng. i did find a way to access to the tree. Not so intuitive, but it works. Idk if there are better way to do that. IJ documentation is horrible

src/main/java/com/redhat/devtools/intellij/common/utils/JSONHelper.java
```diff
@@ -15,14 +15,18 @@
...
 
 import java.io.IOException;
+import java.net.URL;
 import java.util.Map;
 
 public class JSONHelper {
 
     private static final ObjectMapper JSON_MAPPER = new ObjectMapper(new JsonFactory());
 
+    public static JsonNode getJSONFromFile(URL file) throws IOException {
```
jeffmaury: Should be renamed as `getJSONFromURL`

src/main/java/com/redhat/devtools/intellij/common/utils/TreeHelper.java
```diff
@@ -0,0 +1,64 @@
...
+import com.redhat.devtools.intellij.tektoncd.tree.PipelinesNode;
+import com.redhat.devtools.intellij.tektoncd.tree.ResourcesNode;
+import com.redhat.devtools.intellij.tektoncd.tree.TasksNode;
+
+import static com.redhat.devtools.intellij.tektoncd.Constants.*;
+
+public class TreeHelper {
+
+    public static Tree getTree(Project project) {
+        ToolWindow window = ToolWindowManager.getInstance(project).getToolWindow("Tekton");
```
jeffmaury: That could not be part of common as it is Tekton specific. Either move it to tekton packages or add a name parameter

src/main/java/com/redhat/devtools/intellij/tektoncd/completion/DictionaryCompletionProvider.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.io.IOException;
+
+public class DictionaryCompletionProvider extends CompletionProvider<CompletionParameters> {
+    @Override
+    protected void addCompletions(@NotNull CompletionParameters parameters, ProcessingContext context, @NotNull CompletionResultSet result) {
+        try {
+            result.addAllElements(TknDictionary.get().getLookups());
+        } catch (IOException e) {
+            e.printStackTrace();
```
jeffmaury: Should be a call to SLF4J logger

src/main/resources/tknsnippets.json
```diff
@@ -0,0 +1,279 @@
+{
+  "Tekton: ClusterTask": {
+    "prefix": "Tekton: ClusterTask",
+    "body": [
+      "apiVersion: tekton.dev/v1alpha1",
+      "kind: ClusterTask",
+      "metadata:",
+      "  name: ${1:foo}",
+      "  namespace: ${2: namespace}",
```
jeffmaury: Remove the namespace attribute. It is not required when creating a resource and will simplify the code

lstocchi: I thought about this for some time and i added it to keep the **save** consistent with the **update**. When updating, before to push the tekton configuration to the cluster we validate the config by checking some field such as metadata name/namespace and spec. In the current state the **save** works the same. If we remove it we have to pass the namespace through the userData in the virtualFile and we end up having two "problems". 
* First, if the user wants to push to the cluster a tekton resource he previously created outside IJ he can't push it to the cluster without clicking on "new resource" and copy/paste the new content. (In the current situation he can open the file and ctrl+s will work the same and this is also a new feature we can add. When the user click on create "new pipeline" he can choose to open it from the filesystem or create it from scratch).
* Second, I don't think we are going to simplify the code by adding a new userData

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/CreatePipelineAction.java
```diff
@@ -0,0 +1,62 @@
...
+    public CreatePipelineAction() { super(PipelinesNode.class); }
+
+    @Override
+    public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, Tkn tkncli) {
+        String namespace = ((LazyMutableTreeNode)selected).getParent().toString();
+        String content = null;
+        try {
+            content = SnippetHelper.getBody("Tekton: Pipeline");
+        } catch (IOException e) {
+            logger.error("Error: " + e.getLocalizedMessage());
```
jeffmaury: When logging an error the preferred way is `logger.error(e.getLocalizedMessage(), e);` as it allows the stack trace to be logged

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/CreatePipelineAction.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        if (!Strings.isNullOrEmpty(content)) {
+            content = content.replace("${2: namespace}", namespace);
+            Project project = anActionEvent.getProject();
+            VirtualFile fv = ScratchRootType.getInstance().createScratchFile(project, namespace + "-newpipeline.yaml", Language.ANY, content);
+            // append info to the virtualFile to be used during saving
+            fv.putUserData(KIND_PLURAL, "pipelines");
```
jeffmaury: I think there is a constant for that

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/OpenEditorTaskAction.java
```diff
@@ -30,7 +30,7 @@
 import java.io.IOException;
 import java.util.Arrays;
 
-import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_PLURAL;
+import static com.redhat.devtools.intellij.tektoncd.Constants.*;
```
jeffmaury: Use a starred imports is not recommanded (easy to read and managed by IDE)

PR #27, opened 2020-03-18
Added refresh action (#26)

### LINKED ISSUE #26: Add refresh action
We need to add a refresh action so user can explicitly ask to reload the selected node

lstocchi:
it fixes #26


src/main/java/com/redhat/devtools/intellij/common/actions/RefreshAction.java
```diff
@@ -0,0 +1,18 @@
+package com.redhat.devtools.intellij.common.actions;
```
jeffmaury: Add header.

PR #33, opened 2020-03-25
Show logs for pipelinerun/taskrun

### LINKED ISSUE #31: Show logs for pipelinerun/taskrun
Add an action so user can explicitly ask for logs related to a pipelinerun/taskrun and maybe show them during execution

lstocchi:
it fixes #31 

`Show Logs` action is enabled in pipeline/task/pipelinerun/taskrun nodes. When it's used within a pipeline or task node, the user is shown a combobox where he can choose the pipeline/task run he want the logs for.

I had to change the executeWithTerminal method bc it didn't work for me. I could see the new terminal tab created but nothing was displayed and no error occured ..... maybe it was for the IJ version?? I'm using 2019.3


src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/ShowLogsAction.java
```diff
@@ -0,0 +1,106 @@
...
+    public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, Tkn tkncli) {
+        String namespace;
+        String resourceName = selected.toString();
+        List<String> resourceRunsName = null;
+        Class<?> nodeClass = selected.getClass();
+        String kindLabel = "";
+        Notification notification;
+        // if node selected is a Pipeline or a Task we need to find its runs
+        try {
+            namespace = ((LazyMutableTreeNode)selected).getParent().getParent().toString();
```
jeffmaury, line 52: This first part should not be run in the UI thread as it will block IJ if the process is long to complete

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/ShowLogsAction.java
```diff
@@ -0,0 +1,106 @@
...
+            } else if (TaskNode.class.equals(nodeClass)) {
+                List<TaskRun> taskRuns = tkncli.getTaskRuns(namespace, resourceName);
+                if (taskRuns == null || taskRuns.size() == 0) {
+                    throw new IOException("Task " + resourceName + "doesn't have any taskRun to be selected");
+                }
+                resourceRunsName = taskRuns.stream().map(TaskRun::getName).collect(Collectors.toList());
+                nodeClass = TaskRunNode.class;
+                kindLabel = "taskrun";
+            }
+        } catch (IOException e) {
```
jeffmaury, line 70: Switch to Messages

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/ShowLogsAction.java
```diff
@@ -0,0 +1,106 @@
...
+            notification = new Notification(NOTIFICATION_ID, "Error", "An error occurred while requesting logs for " + resourceName + "\n" + e.getLocalizedMessage(), NotificationType.ERROR);
+            Notifications.Bus.notify(notification);
+            logger.error("Error: " + e.getLocalizedMessage(), e);
+            return;
+        }
+
+        if (resourceRunsName != null) {
+            if (resourceRunsName.size() == 1) {
```
jeffmaury, line 78: We need to report an message to user if resourceRunsName is empty (Messages)

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/ShowLogsAction.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        try {
+            if (PipelineRunNode.class.equals(nodeClass)) {
+                tkncli.showLogsPipelineRun(namespace, resourceName);
+            } else if (TaskRunNode.class.equals(nodeClass)) {
+                tkncli.showLogsTaskRun(namespace, resourceName);
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
```
jeffmaury, line 103: Report to the user through Messages

src/main/java/com/redhat/devtools/intellij/tektoncd/ui/component/RunPickerDialog.java
```diff
@@ -0,0 +1,74 @@
...
+        dialog.show();
+        System.exit(0);
+    }
+
+    public String getSelected() {
+        return selected;
+    }
+
+    private void registerListeners() {
+        buttonOK.addMouseListener(new MouseAdapter() {
```
jeffmaury, line 45: I don't understand why you need this code, look at OpenShift dialogs we never used those handlers

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/ShowLogsAction.java
lstocchi: The error is shown inside the try catch above. Isn't that enough? The pipeline/task run lists are checked and if found empty a message is thrown e.g `Pipeline blabla doesn't have any pipelineRun to be selected`

jeffmaury: I did not see the test but I don't think we should report the same way as an error:
- if the user did something wrong I think the dialog should be an information one and not an error one
- don't forget that error dialogs are quite vague and the information may not be quite clear to the user. See https://github.com/redhat-developer/intellij-openshift-connector/blob/master/src/main/java/org/jboss/tools/intellij/openshift/actions/component/LinkServiceAction.java as an example

lstocchi:
> Dare to add FollowLogsAction or delay to another issue ?

Yes i'll make a new PR for this

Issue #34, opened 2020-03-26
Error when resource yaml is wrong


lstocchi:
By using this yaml
```
spec:
  params:
    - name: flags
      type: array
    - name: someURL
      type: string
  steps:
    - name: build
      image: my-builder
      args: ["build", "$(params.flags[*])", "url=$(params.someURL)"]
```

the saving fails with this error 
```
io.fabric8.kubernetes.client.KubernetesClientException: Error while performing the call to https://api.crc.testing:6443/apis/tekton.dev/v1alpha1/namespaces/pipeline-test/tasks/. Response code: 500
	at io.fabric8.kubernetes.client.dsl.internal.RawCustomResourceOperationsImpl.makeCall(RawCustomResourceOperationsImpl.java:582)
	at io.fabric8.kubernetes.client.dsl.internal.RawCustomResourceOperationsImpl.validateAndSubmitRequest(RawCustomResourceOperationsImpl.java:593)
	at io.fabric8.kubernetes.client.dsl.internal.RawCustomResourceOperationsImpl.create(RawCustomResourceOperationsImpl.java:122)
	at com.redhat.devtools.intellij.tektoncd.tkn.TknCli.createCustomResource(TknCli.java:171)
```

Now, by trying it with the terminal i got a clearer message

```
Internal error occurred: admission webhook "webhook.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "params"
```

All of this to say.... we should show an error which is more explanatory

lstocchi:
Another example of this is when there is a non-existent variable
E.g
```
apiVersion: tekton.dev/v1alpha1
kind: ClusterTask
metadata:
  name: foocluster
spec:
  steps:
  - image: ubuntu
  - args:
    - $(inputs.params.ARGS)
    command:
    - /usr/local/bin/tkn
    image: $(inputs.params.tkn-image)
    name: tkn
    resources: {}
```
```
Error from server (InternalError): error when creating "/home/luca/Public/github.com/redhat-developer/intellij-tekton/build/idea-sandbox/config/scratches/newclustertask.yaml": Internal error occurred: admission webhook "webhook.tekton.dev" denied the request: mutation failed: non-existent variable in "$(inputs.params.tkn-image)" for step image: taskspec.steps.image

```

lstocchi:
Closing it as we agreed these are server issues and the plugin only needs to show the error received as it is now

PR #35, opened 2020-03-27
added create task/clustertask/resource actions (#17)

lstocchi:
This is the second PR related to #17. It extends support to create tasks, clustertasks and resources.

I had to change how the push to cluster works. During my testing when i triggered the saving on a specific file, first it tried to save a previously closed file then the one i had opened.

To make it more clear. I had a pipeline with a wrong yaml so the saving was failing miserably (#34). I closed the editor with this file and start to create a new one from scratch. Surprisingly when i wanted to save it, the save process also tried to work with the file i closed previously. Even if it wasn't in my editor windows ... 

To solve this issue i added another control. Now the push action only works on the file which has the focus (selectedEditor).


src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/CreateClusterTaskAction.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, Tkn tkncli) {
+        String content = null;
+        try {
+            content = SnippetHelper.getBody("Tekton: ClusterTask");
+        } catch (IOException e) {
+            logger.error("Error: " + e.getLocalizedMessage(), e);
+        }
+
+        if (!Strings.isNullOrEmpty(content)) {
```
jeffmaury: There is a lot of common code with CreatePipelineAction so I think there is room for refactoring (probably through a common parent class)

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/component/CreateClusterTaskAction.java
```diff
@@ -0,0 +1,58 @@
+/*******************************************************************************
```
jeffmaury: I remember I had once the comment on the package name which seems to be herited from the OpenShift plugin so I think it should be renamed

src/main/java/com/redhat/devtools/intellij/tektoncd/listener/SaveInEditorListener.java
```diff
@@ -143,5 +153,27 @@ public boolean maySaveDocument(@NotNull Document document, boolean isSaveExplici
...
+        // if file is not the one selected, skip it
+        if (selectedEditor.getDocument() != document) return false;
+        // if file is not related to tekton, skip it
+        if (vf == null || vf.getUserData(KIND_PLURAL) == null || vf.getUserData(KIND_PLURAL).isEmpty()) {
+            return false;
+        }
+        return true;
+    }
+
+    private Project getProject() {
```
jeffmaury: There is something buggy here: if you have 2 different projects (windows) opened in IJ, you will get the wrong project and then detect that push is not to be performed. Remember that the Tekton window is attached to a project because the factory is given the Project and then Project can be retrieved when we execute our action as should be stored in the context as well.

src/main/resources/tknsnippets.json
```diff
@@ -6,11 +6,12 @@
       "kind: ClusterTask",
       "metadata:",
       "  name: ${1:foo}",
-      "  namespace: ${2: namespace}",
       "spec:",
       "  steps:",
-      "  - args:",
-      "    - ${3:echo clustertask}"
+      "  - image: ${2:ubuntu}",
```
jeffmaury: Oh man, how can you advertise ubuntu when we have fedora or ubi images

src/main/java/com/redhat/devtools/intellij/tektoncd/listener/SaveInEditorListener.java
```diff
@@ -50,19 +55,16 @@
...
+        if(project == null || !isFileToPush(project, document, vf)) return true;
 
         String namespace, name, apiVersion;
         JsonNode spec;
         CustomResourceDefinitionContext crdContext;
         Notification notification;
         try {
             namespace = YAMLHelper.getStringValueFromYAML(document.getText(), new String[] {"metadata", "namespace"});
-            if (Strings.isNullOrEmpty(namespace)) {
+            if ((namespace == null && !vf.getUserData(KIND_PLURAL).equals(KIND_CLUSTERTASKS)) || namespace == "") {
```
jeffmaury: Note sure to agree with this test as it will allow a namespace for a cluster task

lstocchi:
> We need to properly review base image ? ubi or fedora ?

What do you mean? Make tests with both images to see if there are issues?

jeffmaury:
> 
> 
> > We need to properly review base image ? ubi or fedora ?
> 
> What do you mean? Make tests with both images to see if there are issues when using them?

I mean we sometimes use ubi sometimes fedora we should be uniform and use only one

lstocchi:
> I mean we sometimes use ubi sometimes fedora we should be uniform and use only one

What's the reason behind this? Snippets are useful to show users how a configuration is made and are an example of how one can fill those fields. Showing more different values seems fine to me.

jeffmaury:
> 
> 
> > I mean we sometimes use ubi sometimes fedora we should be uniform and use only one
> 
> What's the reason behind this? Snippets are useful to show users how a configuration is made and are an example of how one can fill those fields. Showing more different values seems fine to me.

Because If I'm a newbie and if I found sometimes ubi sometimes fedora for the same purpose I will be lost.
Cause it send a message to the users and it has a non negligible impact. As I think using ubi will not work on upstream Kubernetes clusters, I would favor switching all to fedora.

lstocchi:
> This will not work on upstream Kubernetes clusters as ubi images are not deployed on docker.io and I'm not sure this image name is valid, did you test it ?

Absolutely not because this is not the purpose of the issue. If we need to make snippets working, the images are just one of the problems here. There are params and resources put as inputs/outputs which are never used in the steps or they have invalid values. Snippets are just valid syntatically so user can have a skeleton to work with. If we need to make snippets valid and usable i suggest to create a tutorial to explain how tekton works and use the pipeline/task/resources used in the tutorial as snippets so newbies can follow the tutorial without writing anything.

jeffmaury:
> 
> 
> > This will not work on upstream Kubernetes clusters as ubi images are not deployed on docker.io and I'm not sure this image name is valid, did you test it ?
> 
> Absolutely not because this is not the purpose of the issue. If we need to make snippets working, the images are just one of the problems here. There are params and resources put as inputs/outputs which are never used in the steps or they have invalid values. Snippets are just valid syntatically so user can have a skeleton to work with. If we need to make snippets valid and usable i suggest to create a tutorial to explain how tekton works and use the pipeline/task/resources used in the tutorial as snippets so newbies can follow the tutorial without writing anything.

I partially agree with you with the exception for the image attribute which is the base mandatory for a task (I mean parameters and resources are optional) and user will expect that default value will be ok.

lstocchi:
FYI ubi is not working, even on openshift. Ubi8/ubi or ubi7/ubi are the correct ones. I switched to fedora so it also works on k8s

PR #41, opened 2020-04-08
Added Follow Logs action

lstocchi:
This PR adds support for `Follow Logs` action


src/main/java/com/redhat/devtools/intellij/tektoncd/actions/logs/LogsBaseAction.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.stream.Collectors;
+
+public class LogsBaseAction extends TektonAction {
+    Logger logger = LoggerFactory.getLogger(LogsBaseAction.class);
+
+    public LogsBaseAction() { super(PipelineRunNode.class, TaskRunNode.class, TaskNode.class, PipelineNode.class); }
+
+    @Override
+    public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, Tkn tkncli) {
+        this.actionPerformed(anActionEvent, path, selected, tkncli);
```
jeffmaury, line 41: Don't understand this ? Infinite loop or dead code ?

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/logs/ShowLogsAction.java
```diff
@@ -0,0 +1,56 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.swing.tree.TreePath;
+import java.io.IOException;
+
+public class ShowLogsAction extends LogsBaseAction {
+    Logger logger = LoggerFactory.getLogger(ShowLogsAction.class);
+
+    @Override
+    public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, Tkn tkncli) {
```
jeffmaury, line 33: Still to much common code. I think the parent class must implement the actionPerformed but calls an abstract method which would be given the node class, the node name and the namespace and would be implemented in each specific action class.

PR #44, opened 2020-04-09
Fix vendor to match marketplace organization

jeffmaury:

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -2,7 +2,7 @@
   <id>com.redhat.devtools.intellij.tekton</id>
   <name>Tekton Pipelines by Red Hat</name>
   <version>1.0</version>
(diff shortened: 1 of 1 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -2,7 +2,7 @@
   <id>com.redhat.devtools.intellij.tekton</id>
   <name>Tekton Pipelines by Red Hat</name>
   <version>1.0</version>
-  <vendor email="support@redhat.com" url="http://www.redhat.com">Red Hat</vendor>
+  <vendor email="support@redhat.com" url="http://www.redhat.com">Red-Hat</vendor>
```
lstocchi, line 5: Are you sure it is Red-Hat? In the OpenShift plugin it is written as Red Hat

jeffmaury: Yes because the new organization stuff in the markeplace does not allow organizations with space in the ID. Once we validate this feature with this plugin, we will update other plugins

PR #51, opened 2020-04-10
Fix for first release - part 1

lstocchi:
@jeffmaury this PR contains almost all changes we discussed yesterday. The only missing one is go from logger.error to logger.warn which i'll cover when this PR will be released.

The timestamp for saving a file works. If you can test it on your IDE it would be great because i don't know why when i click on the terminal my autosave doesn't get triggered.

You told me that validate the model before opening the start dialog wasn't needed right now (and just disable the start btn) but i prefer to go that way because it also helped me solving another issue and it helps having a cleaner code. Also if we need to tackle betav1 soon, this refactor was almost mandatory.
With this, the dialog is not shown if there are no inputs or outputs (the task is started automatically) and an error is shown if the model is not valid. e.g.
error-1

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/common/CommonConstants.java
+++ b/src/main/java/com/redhat/devtools/intellij/common/CommonConstants.java
@@ -17,4 +17,5 @@ public class CommonConstants {
     public static final String HOME_FOLDER = System.getProperty("user.home");
 
     public static final Key<Project> PROJECT = Key.create("project");
+    public static final Key<Long> LAST_MODIFICATION_STAMP = Key.create("last.modification.stamp");
 }
--- a/src/main/java/com/redhat/devtools/intellij/common/CommonConstants.java
+++ b/src/main/java/com/redhat/devtools/intellij/common/CommonConstants.java
@@ -16,6 +16,6 @@ import com.intellij.openapi.util.Key;
 public class CommonConstants {
     public static final String HOME_FOLDER = System.getProperty("user.home");
 
-    public static final Key<Project> PROJECT = Key.create("project");
-    public static final Key<Long> LAST_MODIFICATION_STAMP = Key.create("last.modification.stamp");
+    public static final Key<Project> PROJECT = Key.create("com.redhat.devtools.intellij.common.project");
+    public static final Key<Long> LAST_MODIFICATION_STAMP = Key.create("com.redhat.devtools.intellij.common.last.modification.stamp");
 }
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/StartAction.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/StartAction.java
@@ -29,6 +29,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.swing.tree.TreePath;
 import java.io.IOException;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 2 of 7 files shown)


src/main/java/com/redhat/devtools/intellij/tektoncd/actions/StartAction.java
```diff
@@ -60,18 +62,32 @@ public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object s
...
-            if (stdialog.isOK()) {
+
+            StartResourceModel model = new StartResourceModel(configuration, resources);
+
+            if (!model.isValid()) {
+                UIHelper.executeInUI(() -> Messages.showErrorDialog(model.getErrorMessage(), "Error"));
+                return;
+            }
+
+            boolean noInputsAndOuputs = model.getInputs() == null && model.getOutputs() == null;
```
jeffmaury, line 73: Just for clarity I think it's better if inputs and outputs are always there (never null) but check should be isEmpty

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/StartAction.java
```diff
@@ -60,18 +62,32 @@ public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object s
...
+            if (!noInputsAndOuputs) {
+                stdialog = UIHelper.executeInUI(() -> {
+                    StartDialog dialog = new StartDialog(null, model.getNamespace(), model.getName(), model.getKind(), model.getInputs(), model.getOutputs(), resources);
+                    dialog.show();
+                    return dialog;
+                });
+            }
+            if (noInputsAndOuputs || stdialog.isOK()) {
                 try {
+                    Map<String, String> params = stdialog == null ? null : stdialog.getParameters();
```
jeffmaury, line 85: In order to respect the MVC model, dialog should be given the model and modify it so once dialog is closed you should work on the model

src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/component/field/Input.java
```diff
@@ -80,6 +86,16 @@ public Input fromJson(JsonNode inputNode, Kind kind) {
         return this;
     }
 
+    private String convertArrayNodeToString(ArrayNode node) {
+        String result = "";
+        for (Iterator<JsonNode> it = node.iterator(); it.hasNext(); ) {
```
jeffmaury, line 91: Could be rewritten with streams and Collectors.join

src/main/java/com/redhat/devtools/intellij/common/CommonConstants.java
```diff
@@ -17,4 +17,5 @@
     public static final String HOME_FOLDER = System.getProperty("user.home");
 
     public static final Key<Project> PROJECT = Key.create("project");
+    public static final Key<Long> LAST_MODIFICATION_STAMP = Key.create("last.modification.stamp");
```
jeffmaury, line 20: After the fact, I realize that this must be namespaced to make sure other plugins may not overrwrite those value. So should be prefixed by com.redhat.devtools.intellij.common

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/StartAction.java
lstocchi: postponed to #53

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/StartAction.java
lstocchi: postponed to #53

src/main/java/com/redhat/devtools/intellij/tektoncd/listener/SaveInEditorListener.java
```diff
@@ -56,7 +57,15 @@
     public boolean maySaveDocument(@NotNull Document document, boolean isSaveExplicit) {
         VirtualFile vf = FileDocumentManager.getInstance().getFile(document);
         Project project = vf.getUserData(PROJECT);
-        if(project == null || !isFileToPush(project, document, vf)) return true;
+        Long lastModificationStamp = vf.getUserData(LAST_MODIFICATION_STAMP);
+        Long currentModificationStamp = document.getModificationStamp();
+        if(project == null ||
+                 !isFileToPush(project, document, vf) ||
+                 (lastModificationStamp != null && currentModificationStamp.equals(lastModificationStamp))
```
jeffmaury, line 64: if you're using equals then you don't need to null check

PR #55, opened 2020-04-10
Change root node icon (Tekton) and add Tekton icon to the window

jeffmaury:


src/main/resources/META-INF/plugin.xml
```diff
@@ -26,7 +26,7 @@
     <fileDocumentSynchronizationVetoer implementation="com.redhat.devtools.intellij.tektoncd.listener.SaveInEditorListener" order="first" />
     <completion.contributor language="any" implementationClass="com.redhat.devtools.intellij.tektoncd.completion.DictionaryContributor"/>
     <fileDocumentSynchronizationVetoer implementation="com.redhat.devtools.intellij.tektoncd.listener.SaveInEditorListener" order="first"/>
-    <toolWindow id="Tekton" anchor="left" factoryClass="com.redhat.devtools.intellij.tektoncd.WindowToolFactory"/>
+    <toolWindow id="Tekton" anchor="left" factoryClass="com.redhat.devtools.intellij.tektoncd.WindowToolFactory" icon="/images/cluster.png"/>
```
lstocchi, line 29: Not sure this is the correct icon. Why don't we use the tekton one? This is the OpenShift icon

jeffmaury: I changed it with this PR

PR #57, opened 2020-04-11
Add start and completion time text next to tree node

### LINKED ISSUE #54: Runs should be displayed with start and end times
See this example:
image

lstocchi:
it fixes #54 

image


lstocchi:
image

src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/PipelineRunDeserializer.java
```diff
@@ -36,6 +36,16 @@ public PipelineRunDeserializer() {
                 String name = item.get("metadata").get("name").asText();
                 Optional<Boolean> completed = Optional.empty();
                 JsonNode conditions = item.get("status").get("conditions");
+                String startTimeText = "running " + DateHelper.humanizeDate(item.get("status").get("startTime").asText());
```
jeffmaury, line 39: I don't agree. You are mixing the model and the UI. The model object should only store the date and the rendering should decided to generate the label based on different data (status, dates). Also I would stored the dates as object (Instant, Calendar, long) instead of text but the Fabric8 API uses strings so....

lstocchi:
When working on fixing the issues and moving the creation of the label to the ui i noticed that pipelinerun and taskrun are identical. I refactored it a bit. Give it a look @jeffmaury

src/main/java/com/redhat/devtools/intellij/common/utils/DateHelper.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Contributors:
+ * Red Hat, Inc.
+ ******************************************************************************/
+package com.redhat.devtools.intellij.common.utils;
+
+import java.time.Duration;
+import java.time.Instant;
+
+public class DateHelper {
```
jeffmaury, line 16: Would like to see a unit test with various combinations

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/logs/LogsBaseAction.java
```diff
@@ -44,12 +44,30 @@ public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object s
...
 
-            this.actionPerformed(namespace, resourceName, nodeClass, tkncli);
+            this.actionPerformed(namespace, cleanName(resourceName), kind, nodeClass, tkncli);
         });
     }
 
-    public abstract void actionPerformed(String namespace, String resourceName, Class nodeClass, Tkn tkncli);
+    public abstract void actionPerformed(String namespace, String resourceName, String kind, Class nodeClass, Tkn tkncli);
+
+    private String cleanName(String resourceName) {
```
jeffmaury, line 55: No your are using an army: the node should have a name and if the visual representation is different then either you overload the toString method or you update the cell renderer.

lstocchi:
Added the steps under the pipelinerun but i have a question. 
This is what we have vs vscode
run-example
The taskRuns under the task are displayed better in IJ for me. In vscode there are empty nodes or even if they have the name there is no way to understand which pipeline triggered them. You have to open their YAML.
The problem is under the pipelineRun. In vscode they display the name of the step which i admit could be useful. Which way should we go? It's possible to also have steps name under pipelinerun and taskrun name under task. WDYT? @jeffmaury

jeffmaury:
> 
> 
> Added the steps under the pipelinerun but i have a question.
> This is what we have vs vscode
> run-example
> The taskRuns under the task are displayed better in IJ for me. In vscode there are empty nodes or even if they have the name there is no way to understand which pipeline triggered them. You have to open their YAML.
> The problem is under the pipelineRun. In vscode they display the name of the step which i admit could be useful. It's possible to also have steps name under pipelinerun and taskrun name under task. WDYT? @jeffmaury

Their display for taskRuns is not that stupid. Instead of displaying the taskRun name (which is of little value), they display the pipeline that causes this taskRun to start). The only issue is that the display when no pipeline causes the taskrun to start is a little bit strange.

lstocchi:
> Their display for taskRuns is not that stupid. Instead of displaying the taskRun name (which is of little value), they display the pipeline that causes this taskRun to start). The only issue is that the display when no pipeline causes the taskrun to start is a little bit strange.

I don't agree much. If we take the example above `test-release` is a task and below it there are the taskRuns. They only display the name of the step (as it is written in the pipeline) where the task has been run but they don't display the pipeline name at all. I only see that the taskRun named `first` was executed successfully but there is no info regarding the pipeline. I have to remember exactly which pipeline that step belongs. In opposite the taskRun name gives info regarding which is the pipeline that launched it.

jeffmaury:
> 
> 
> > Their display for taskRuns is not that stupid. Instead of displaying the taskRun name (which is of little value), they display the pipeline that causes this taskRun to start). The only issue is that the display when no pipeline causes the taskrun to start is a little bit strange.
> 
> I don't agree much. If we take the example above `test-release` is a task and below it there are the taskRuns. They only display the name of the step (as it is written in the pipeline) where the task has been run but they don't display the pipeline name at all. I only see that the taskRun named `first` was executed successfully but there is no info regarding the pipeline. I have to remember exactly which pipeline that step belongs. In opposite the taskRun name gives info regarding which is the pipeline that launched it.

You're right, in my case the pipeline, step and task were having the same name. So can't we be smarted and display `pipeline_name/step_name` as a prefix ?

lstocchi:
> You're right, in my case the pipeline, step and task were having the same name. So can't we be smarter and display `pipeline_name/step_name` as a prefix ?

My only concern is the length of pipeline_name and/or the step_name. By making a quick test i found out that in the taskrun name their values are cut if too long.
So i would say to do the same thing. 
So in our tree we would see the taskRun name in 3 different ways:
1) under the pipelineRun -> task_name/step_name
2) under the task (if the taskRun was created by runnung a pipeline) -> pipeline_name/step_name
3) under the task (the taskRun was created by running the task itself) -> taskRun name

Below an example of long step_name cut
image

jeffmaury:
> 
> 
> > You're right, in my case the pipeline, step and task were having the same name. So can't we be smarter and display `pipeline_name/step_name` as a prefix ?
> 
> My only concern is the length of pipeline_name and/or the step_name. By making a quick test i found out that in the taskrun name their values are cut if too long.
> So i would say to do the same thing.
> So in our tree we would see the taskRun name in 3 different ways:
> 
>     1. under the pipelineRun -> task_name/step_name
> 
>     2. under the task (if the taskRun was created by runnung a pipeline) -> pipeline_name/step_name
> 
>     3. under the task (the taskRun was created by running the task itself) -> taskRun name
> 
> 
> Below an example of long step_name cut
> image

My proposition is for 3 not to display the name but only the start and end times. I had another idea: when a task run is run from a pipeline, don't show it in the tasks nodes but rather in the pipelines nodes. But that would need to be checked with Siamak. So for your release we could implement a limit algorithm that would limit the pipeline + step to be under 16 chars for example

lstocchi:
image

lstocchi:
image

Issue #72, opened 2020-04-21
Change icons to make this plugin looking like VSCode extension


lstocchi:
There are multiple icons which are different between this plugin and vscode extension. We should use the same set of icons in both.

PR #90, opened 2020-04-27
Added Pipelineruns and Taskruns node as root

### LINKED ISSUE #81: Add PipelineRuns and TaskRuns in the tree as root nodes
PipelineRuns can be standalone without referencing a pipeline and should be displayed in their own folder, in addition to under a particular pipeline

lstocchi:
it fixes #81 

Couple of question here. 
First is which icon i should use? I left the pipeline icon as default here because i opened https://github.com/redhat-developer/intellij-tekton/issues/72 and i'll move to different icons with that PR. Do we agree to use the icons from vscode?

Second, in vscode under the pipelineruns root node they only show the pipelineruns without being able to expand them and see their taskrun. In this PR it works differently and you can expand a pipelinerun and see its children. Do i have to change it?

image

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/tree/ParentableNode.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/tree/ParentableNode.java
@@ -32,4 +32,16 @@ public abstract class ParentableNode<T> {
     public String getName() {
         return name;
     }
+
+    public String getNamespace() {
+        Object element = this;
+        while (!(element instanceof NamespaceNode)) {
+            if (element instanceof ParentableNode) {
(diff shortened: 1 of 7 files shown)


src/main/java/com/redhat/devtools/intellij/tektoncd/tree/ParentableNode.java
```diff
@@ -32,4 +32,16 @@ public T getParent() {
     public String getName() {
         return name;
     }
+
```
jeffmaury, line 35: This is formally incorrect as it introduce a circular dependency between the 2 classes but we can keep it

Issue #89, opened 2020-04-27
Support multiple resources creation from one single file


lstocchi:
In many tutorials all the pipelines/tasks/resources used, are taken/created by a single file that contains them all. We should allow users to do the same in intellij, create multiple resources from a single file

PR #94, opened 2020-04-29
Added support for Conditions

### LINKED ISSUE #75: Add support for conditions
[Tekton conditions](https://github.com/tektoncd/pipeline/blob/master/docs/conditions.md) enable conditional execution of tasks in a pipeline. Conditions should be added in the Tekton window in the tree to allow user to view and edit them.

lstocchi:
it resolves #75 

An example of tree below. It shows the actual task and as its children all the condition checks. The strange fact is that if you have multiple condition of the same type (e.g 2 `is-equal` condition) tekton pipelinerun only shows the result of both of them. So you don't actually know which one failed and the parameters that made it failed. Maybe we need to open an issue on Tekton or ask Siamak.
Other thing is why the pipelinerun is shown successfull if the task under it has been never executed.... 

image

As a description of the error i used to show the `reason` because the message in the pipelinerun is too long and it doesn't give any additional info (e.g `ConditionChecks failed for Task condition-pipeline-run-54c7c-process-g2m5m in PipelineRun condition-pipeline-run-54c7c`)


jeffmaury:
Sorry seems my patch for the UI thread issue cause this to be rebased, can you doit before I review ?

lstocchi:
Yeah I didn't update the TreeHelper because the refresh wasn't working anymore for me and i wanted to investigate it. I mean it starts working only in debug otherwise sometimes it works sometimes it doesn't. Opened an issue here https://github.com/redhat-developer/intellij-tekton/issues/102 . Don't you have any problem in Windows?
BTW i added the missing part in TreeHelper. 

Regarding the icons, an issue is opened https://github.com/redhat-developer/intellij-tekton/issues/72 . I was waiting your confirmation to use vscode icons :+1:

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/TreeHelper.java
```diff
@@ -75,6 +77,7 @@ public static Class retrieveNodeClassByKind(String kind) {
             case KIND_PIPELINES: return PipelinesNode.class;
             case KIND_RESOURCES: return ResourcesNode.class;
             case KIND_TASKS: return TasksNode.class;
+            case KIND_CONDITIONS: return ConditionsNode.class;
```
jeffmaury, line 80: Aren't we missing TaskRuns as well now ?

lstocchi: Not now. This refresh is only called from the saveListener and it is only triggered from virtual files created by the plugin because of their hidden user data (so pipeline, task, clustetask, resources and condition). Taskruns are only created by starting a task. We need to add taskruns support when adding this https://github.com/redhat-developer/intellij-tekton/issues/89

PR #95, opened 2020-04-30
Activate completion code only on YAML files (#93)

### LINKED ISSUE #93: Snippets are enabled for any language
That causes code assist on Java files quite unusable

lstocchi:
it resolves #93 

Completion code is only enabled on YAML file

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -29,6 +29,7 @@
        on how to target different products -->
   <depends>com.intellij.modules.lang</depends>
   <depends>org.jetbrains.plugins.terminal</depends>
+  <depends>org.jetbrains.plugins.yaml</depends>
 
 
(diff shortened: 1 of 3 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -29,6 +29,7 @@
        on how to target different products -->
   <depends>com.intellij.modules.lang</depends>
   <depends>org.jetbrains.plugins.terminal</depends>
+  <depends>org.jetbrains.plugins.yaml</depends>
```
jeffmaury, line 32: I would have expected that line 38 would be updated as well, no ?

lstocchi: No, i tried changing it to yaml but it makes the completion stop working. At first i thought that that was the only option to be changed to make it works with yaml file but no... i really don't know how it works or maybe yaml language has some issue

PR #97, opened 2020-04-30
Added support for serviceAccount (#68)

### LINKED ISSUE #68: Support serviceAccount when starting pipeline/task

lstocchi:
it resolves #68

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/utils/StartResourceModel.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/utils/StartResourceModel.java
@@ -79,7 +79,9 @@ public class StartResourceModel {
         }
         if (isPipeline()) {
             saNode = YAMLHelper.getValueFromYAML(configuration, new String[] {"spec", "serviceAccountNames"});
-            getTaskServiceAccountFromNode(saNode);
+            if (saNode != null) {
+                getTaskServiceAccountFromNode(saNode);
+            }
         }
     }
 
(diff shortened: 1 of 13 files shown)


lstocchi:
@jeffmaury added combo with serviceAccounts taken from the cluster

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/StartResourceModel.java
```diff
@@ -147,7 +176,7 @@ private void setDefaultValueResources() {
             result.addAll(getInputsFromNodeInternal(resources, Input.Kind.RESOURCE));
         }
 
-        return result.isEmpty() ? null : result;
+        return result.isEmpty() ? Collections.emptyList() : result;
```
jeffmaury, line 179: Return result directly

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/StartResourceModel.java
```diff
@@ -170,7 +199,7 @@ private void setDefaultValueResources() {
             }
         }
 
-        return result.isEmpty() ? null : result;
+        return result.isEmpty() ? Collections.emptyList() : result;
```
jeffmaury, line 202: Return result directly

jeffmaury:
As you updated the model constructor, you must update the corresponding test and see if there are new test methods to be added as well

lstocchi:
> As you updated the model constructor, you must update the corresponding test and see if there are new test methods to be added as well

Yes i added a couple of tests to verify if serviceName/taskServiceNames are correctly filled in when using a pipeline vs a task. Let me know if you think we need something else

PR #101, opened 2020-05-07
added support for triggers (#74)

### LINKED ISSUE #74: Add support for trigger resource types
[Tekton Triggers](https://github.com/tektoncd/triggers) enable creating webhooks for pipelines and include the following CRs:

- TriggerTemplate
- TriggerBinding
- ClusterTriggerBinding
- EventListener

The trigger CRD should be displayed in the Tekton window

lstocchi:
it resolves #74

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCli.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCli.java
@@ -184,7 +184,7 @@ public class TknCli implements Tkn {
 
     @Override
     public String getEventListenerYAML(String namespace, String eventListener) throws IOException {
-        return ExecHelper.execute(command, "eventListener", "describe", eventListener, "-n", namespace, "-o", "yaml");
+        return ExecHelper.execute(command, "eventlistener", "describe", eventListener, "-n", namespace, "-o", "yaml");
     }
 
     @Override
@@ -224,7 +224,7 @@ public class TknCli implements Tkn {
 
     @Override
     public void deleteEventListener(String namespace, String eventListener) throws IOException {
-        ExecHelper.execute(command, "eventListener", "delete", "-f", eventListener, "-n", namespace);
+        ExecHelper.execute(command, "eventlistener", "delete", "-f", eventListener, "-n", namespace);
     }
 
     @Override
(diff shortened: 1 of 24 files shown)


src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCli.java
```diff
@@ -69,6 +68,11 @@ public boolean isTektonAware(KubernetesClient client) {
         return client.rootPaths().getPaths().stream().filter(path -> path.endsWith("tekton.dev")).findFirst().isPresent();
     }
 
+    @Override
+    public boolean isTektonTriggersAware(KubernetesClient client) {
+        return client.rootPaths().getPaths().stream().filter(path -> path.endsWith("triggers.tekton.dev")).findFirst().isPresent();
```
jeffmaury, line 73: On only issue if that if you have triggers installed but not pipelines, then the `isTektonAware` will return true but I'm not sure this is relevant

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/CRDHelper.java
```diff
@@ -13,17 +13,32 @@
...
+                .withGroup(group)
                 .withScope("Namespaced")
-                .withVersion(apiVersion[1])
+                .withVersion(version)
                 .withPlural(plural)
                 .build();
     }
+
+    public static boolean isClusterScopedResource(String kind_plural) {
+        switch (kind_plural) {
```
jeffmaury, line 36: Could be rewritten as `return kind_plural == KIND_CLUSTERTASKS || kind_plural == KIND_CLUSTERTRIGGERBINDINGS;`

src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCli.java
lstocchi: I didn't think at this because in the documentation they state you need pipeline before to install triggers so i thought there were some kind of prereq and if you didn't have one you coldn't install the other. I wanted to make a test by uninstalling everything from my cluster but it's like the uninstall process doesn't remove everything or my cluster had some issue. The plugin still see tekton installed. Opening an issue to investigate more about this because if the uninstall doesn't remove the tekton.dev path and we check that to see if tekton is installed we could have problem.

PR #107, opened 2020-05-11
Updated icons (#72)

### LINKED ISSUE #72: Change icons to make this plugin looking like VSCode extension

lstocchi:
it resolves #72 

image


adietish:
Consistency is a great. On the other hand I think that these icons aren't ideal in the first place: The added value of icons is that you instantly grock what it's all about similar to "a picture is worth 1000 words". These icons lack this power, looking at them is not good enough, you have to read them to understand their meaning. Shouldn't we look for proper images rather than "iconized labels"?

lstocchi:
I agree. I don't like these icons as well. The problem is that it is difficult to find icons that add some value in this case but we are planning to talk to a PM and UX guy and we can ask them to handle this.

In the short term i think we need this PR because we added many other nodes (taskrun/conditions/triggers...) and we don't have a proper icons for them (i mean they don't exist at all, the old ones were taken from vscode but they switched to these iconized labels) so the only solution is to reuse the pipeline or task icons but it can create some confusion.

adietish:
@lstocchi ok, I unerstand the reason for having these (consistency and lack of icons), my point is that we maybe should bring in proper icons for both IDEs (I'm not complaining about this PR specifically)

lstocchi:
> I don't see the PNG files being deleted, any reason ?

No, just forgot to remove them. But i also need to update the code when the triggers PR will be merged so i'll fix this as well

PR #113, opened 2020-05-13
Fix model to report several output resources

### LINKED ISSUE #112: When a task is defined with several output resources only one is displayed when starting the task

jeffmaury:
Fixes #112


src/main/java/com/redhat/devtools/intellij/tektoncd/utils/StartResourceModel.java
```diff
@@ -166,7 +168,9 @@ private void setDefaultValueResources() {
         if (resources != null) {
             for (Iterator<JsonNode> it = resources.iterator(); it.hasNext(); ) {
```
lstocchi, line 169: It's my fault so i could also rewrite this part later if you want.
 
`List<JsonNode> resources = outputsNode.findValues("resources");` doesn't make sense here. There is just one resources section so we can better modify it to `JsonNode resources = outputsNode.get("resources")` and then avoid making two loops and end with 
```
        if (resources != null) {
            for (Iterator<JsonNode> it = resources.elements(); it.hasNext(); ) {                
                    result.add(new Output().fromJson(it.next()));                
            }
        }
```

PR #115, opened 2020-05-18
Add support for v1beta1 (#78)

### LINKED ISSUE #78: Add support for v1beta1 API
Update APIs to support v1beta1 and also remove deprecated attributes from snippets.

lstocchi:
it resolves #78

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/utils/CRDHelper.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/utils/CRDHelper.java
@@ -20,9 +20,11 @@ public class CRDHelper {
     public static CustomResourceDefinitionContext getCRDContext(String apiVersion, String plural) {
         if (Strings.isNullOrEmpty(apiVersion) || Strings.isNullOrEmpty(plural)) return null;
         String[] groupVersion = apiVersion.split("/");
-        if (groupVersion.length != 2) return null;
+        if (groupVersion.length != 2 && !apiVersion.equalsIgnoreCase("triggers.tekton.dev")) {
+            return null;
+        }
         String group = groupVersion[0];
-        String version = groupVersion[1];
+        String version = groupVersion.length > 1 ? groupVersion[1]: "v1alpha1";
         return new CustomResourceDefinitionContext.Builder()
                 .withName(plural + "." + group)
                 .withGroup(group)
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -16,8 +16,8 @@
   <p><b>0.0.1</b></p>
   <p>Initial release</p>
   <p>Allow create, edit and start actions for pipelines, tasks and resources</p>
-  <p><b>WARNING</b>: this plugin has been tested and validated against Tekton Pipelines Alpha1.
-  You may have some issues against Tekton Pipelines Beta1</b>
+  <p><b>WARNING</b>: this plugin has been tested and validated against Tekton Pipelines Beta1.
+  You may have some issues against Tekton Pipelines Alpha1</b>
     ]]>
   </change-notes>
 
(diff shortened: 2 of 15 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -16,8 +16,8 @@
   <p><b>0.0.1</b></p>
   <p>Initial release</p>
   <p>Allow create, edit and start actions for pipelines, tasks and resources</p>
-  <p><b>WARNING</b>: this plugin has been tested and validated against Tekton Pipelines Alpha1.
-  You may have some issues against Tekton Pipelines Beta1</b>
+  <p><b>WARNING</b>: this plugin has been tested and validated against Tekton Pipelines Beta1.
+  You may have some issues against Tekton Pipelines Alpha1</b>
```
jeffmaury, line 20: We should be more strict for Alpha1

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/CRDHelper.java
```diff
@@ -20,9 +20,11 @@
     public static CustomResourceDefinitionContext getCRDContext(String apiVersion, String plural) {
         if (Strings.isNullOrEmpty(apiVersion) || Strings.isNullOrEmpty(plural)) return null;
         String[] groupVersion = apiVersion.split("/");
-        if (groupVersion.length != 2) return null;
+        if (groupVersion.length != 2 && !apiVersion.equalsIgnoreCase("triggers.tekton.dev")) {
```
jeffmaury, line 23: I don't understand this test ? Does it mean this code is only relevant for trigger related resources ?

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/CRDHelper.java
```diff
@@ -20,9 +20,11 @@
...
+            return null;
+        }
         String group = groupVersion[0];
-        String version = groupVersion[1];
+        String version = groupVersion.length > 1 ? groupVersion[1]: "v1alpha1";
```
jeffmaury, line 27: You claimed we switched to v1beta1 so shouldn't it be v1beta1 ?

lstocchi: Yes but not all resources switched to beta, only task, taskrun, pipeline, pipelinerun. Resources are still and always be alpha because they will be removed soon. Conditions are still alpha bc they are new. Triggers still alpha, i guess for the same reason as conditions. 

I added this control because any resource in beta will explicitly declare that in its version (otherwise we are not sure which one to use alpha or beta), but if it's not explicitly declared it will be alpha for sure. An example are triggers. Idk the behavior in minishift but in CRC triggers will have version "triggers.tekton.dev". Even if you add a resource using "triggers.tekton.dev/v1alpha1" it will be renamed to "triggers.tekton.dev"

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/CRDHelper.java
lstocchi: Reason in comment below. Let me know what you think

src/main/resources/META-INF/plugin.xml
lstocchi: What do you mean?

jeffmaury: We should claim this plugin requires Pipeline Beta1 as a minimum.

lstocchi: Ah ok. Yeah it makes sense. I guess i'm going to revert this change and open a new issue then. In the change-notes we should also add that you need to install triggers if you want to use them. I see all of this together when we're going to add the new plugin-version in the change-notes. I also discovered a user can install a specific version of a plugin so it makes sense to leave this warning message as it is for the first version and add a new one with the next release

lstocchi: Opened https://github.com/redhat-developer/intellij-tekton/issues/118

PR #116, opened 2020-05-18
Change way namespace is retrieved when pushing resource to cluster

### LINKED ISSUE #86: Stop asking user to add namespace in config
Forcing the user to specify the namespace is annoying and counter productive because most of the samples don't specifiy it. We should save it in the callback context and retrieve it when we push to the cluster.

lstocchi:
it resolves #86


lstocchi:
@jeffmaury when i was doing tests i noticed that we never implemented basic stuff for clustertask so i added it here to do further tests with clusterscoped resources.

PR #122, opened 2020-06-04
Open tekton resource in editor on tekton tree item double-click

### LINKED ISSUE #32: Double-click on a Tekton resource should open its YAML
When user double-clicks on a Tekton resource the tree gets expanded/collapsed. We should also open the resource YAML in the editor.

evidolob:
Fix #32

This PR add own Double click handler for tree, it allow to handle double click event for any node type: leaf and not leaf. And it unify editor opening with `Open in Editor` action and double click handler.

Demo:
ezgif com-video-to-gif (21)


src/main/java/com/redhat/devtools/intellij/tektoncd/actions/OpenEditorAction.java
```diff
@@ -16,87 +16,44 @@
...
-import com.redhat.devtools.intellij.tektoncd.tree.ClusterTriggerBindingNode;
-import com.redhat.devtools.intellij.tektoncd.tree.EventListenerNode;
-import com.redhat.devtools.intellij.tektoncd.tree.NamespaceNode;
-import com.redhat.devtools.intellij.tektoncd.tree.ParentableNode;
-import com.redhat.devtools.intellij.tektoncd.tree.PipelineNode;
-import com.redhat.devtools.intellij.tektoncd.tree.ResourceNode;
-import com.redhat.devtools.intellij.tektoncd.tree.TaskNode;
-import com.redhat.devtools.intellij.tektoncd.tree.TriggerBindingNode;
-import com.redhat.devtools.intellij.tektoncd.tree.TriggerTemplateNode;
+import com.redhat.devtools.intellij.tektoncd.tree.*;
```
lstocchi, line 22: We don't use wildcard imports. I suggest you to disable intellij autoimport otherwise you will end up with this issue often https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/OpenEditorAction.java
```diff
@@ -16,87 +16,44 @@
...
             UIHelper.executeInUI(() -> Messages.showErrorDialog("Error: " + e.getLocalizedMessage(), "Error"));
         }
 
-        if (!content.isEmpty()) {
+        if (yamlAndKind != null && !yamlAndKind.first.isEmpty()) {
             Project project = anActionEvent.getProject();
 
             Optional<FileEditor> editor = Arrays.stream(FileEditorManager.getInstance(project).getAllEditors()).
-                    filter(fileEditor -> fileEditor.getFile().getName().startsWith(namespace + "-" + element.getName()) &&
+                    filter(fileEditor -> fileEditor.getFile().getName().startsWith(namespace + "-" + element.getName() + ".yaml") &&
```
lstocchi, line 53: I guess this can be edited in equals() and `fileEditor.getFile().getExtension().equals("yaml")` can be removed if we explicitly declare the extension in the first check. I guess we didn't add the ".yaml" because we were using another way to display resources and a number got appended (e.g "res_1.yaml, res_2.yaml") but we switched to temp file.

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/OpenEditorAction.java
```diff
@@ -16,87 +16,44 @@
...
                             fileEditor.getFile().getExtension().equals("yaml")).findFirst();
             if (!editor.isPresent()) {
-                createAndOpenVirtualFile(project, namespace, namespace + "-" + element.getName() + ".yaml", content, kind, null);
+                VirtualDocumentHelper.createAndOpenVirtualFile(project, namespace, namespace + "-" + element.getName() + ".yaml", yamlAndKind.first, yamlAndKind.second);
```
lstocchi, line 56: I would keep only one method of `createAndOpenVirtualFile`. Maybe we should delete the one inside TektonAction (as they are the same) and update the one inside VirtualDocumentHelper to make it accepts the extra parameters used to pass the targetnode.

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/OpenEditorAction.java
```diff
@@ -16,87 +16,44 @@
...
-    public OpenEditorAction() { super(TaskNode.class, PipelineNode.class, ResourceNode.class, ClusterTaskNode.class, ConditionNode.class, TriggerTemplateNode.class, TriggerBindingNode.class, ClusterTriggerBindingNode.class, EventListenerNode.class); }
+    public OpenEditorAction() {
+        super(TaskNode.class, PipelineNode.class, ResourceNode.class, ClusterTaskNode.class, ConditionNode.class, TriggerTemplateNode.class, TriggerBindingNode.class, ClusterTriggerBindingNode.class, EventListenerNode.class);
+    }
 
 
     @Override
     public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, Tkn tkncli) {
         ParentableNode<? extends ParentableNode<NamespaceNode>> element = getElement(selected);
         String content = "";
```
lstocchi, line 40: `content` can be deleted.

src/main/java/com/redhat/devtools/intellij/tektoncd/tree/TektonTree.java
```diff
@@ -0,0 +1,104 @@
...
+
+        public TektonNavigatable(ParentableNode<?> node, String content, String kind) {
+            this.node = node;
+            this.content = content;
+            this.kind = kind;
+        }
+
+        @Override
+        public void navigate(boolean requestFocus) {
+            Project project = node.getRoot().getProject();
```
lstocchi, line 80: This is the same code used inside the openInEditor action. We could use a common method.

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/OpenEditorAction.java
evidolob: Yes, that may bad, I wasn't change default settings for wildcard imports, going to fix that

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/OpenEditorAction.java
evidolob: OK, I change that, because I face problem when I have two pipeline which starts with same name, I wasn't able to open second(`foo.yam`), with longer name if first was open(`fooBar.yam`).

evidolob:
> The first is that the double click only opens an editor if the node is a leaf otherwise it will expand the children. Do we want this behavior or it should always open an editor (I don't know if it's too hard to achieve/if it's possible at all)?

I don't think that possible, at list with `EditSourceOnDoubleClickHandler`, I think we can do it manually, by counting click events, but it may confuse if you want just expand/collapse node, you make double click and editor opens.

> Second is, the `getData` method inside `TektonTree` seems to always fire. By looking at the `EditSourceOnDoubleClickHandler.install` i saw it also accepts a runnable as argument and it will be fired (if not null) when the event is triggered. Couldn't we use that and maybe retrieved the node we need by checking the current selected node in the tree? This is just an idea as i don't know how easy it is to do. IJ is always a pain.

That was my first plan, but I look on how IJ use that `EditSourceOnDoubleClickHandler`, and internally they use `Navigatable ` and `getData` to get it. I just want to do same as rest platform does. I'm not very happy with current implementation, it more complicated that it should be.
I will mark this PR as draft, and rework it.

@lstocchi Thx for review.

evidolob:
> Yes i agree but then i start thinking.. is it the double click more useful to open in editor or to expand a node? From a user perspective, the expand/collapse is already a single click (on the left of the node but still a single click) while the open in editor is a right click + a left click on the action. So maybe there is a bigger advantage if we open an editor.

Let do next, I update this PR with new implementation of opening file on double click(current behaviour, open editor only for leaf nodes) and then prepare alternative PR (if it possible) with opening on every node where it possible. My concern there, that we may introduce completely new tree behaviour in terms of UX, I cannot find any other tree in IJ thad do the same. This also may be confusing.

jeffmaury:
I found one: in the structure panel, if you double click on the root node (the class), it will both expand/collapse the tree and open the Java file. I agree that double click should operate on non leaf nodes as certains represents Tekton resources (ex pipelines)

evidolob:
@jeffmaury In my IJ it open Java file on single click

jeffmaury:
Not my behavior I had to install a plugin for that but then removed it

evidolob:
I use latest IJ, structure panel has option:
Screenshot 2020-06-05 at 10 07 54
We may consider similar behaviour

evidolob:
@lstocchi @jeffmaury 
I found that we can change number of clicks to expand/collapse node, if I change it to 1 then tree behaves as this:
ezgif com-video-to-gif (20)

Only thing there is when we double click on item, it expands/collapse twice before editor is opened.

lstocchi:
I would prefer to do `expand/collapse` and `open in editor` at the same time. The fact that if you double click the `expand/collapse` is executed twice it's annoying.

evidolob:
`Tree` class has `setToggleClickCount` method.

>I would prefer to do expand/collapse and open in editor at the same time. The fact that if you double click the expand/collapse is executed twice it's annoying.

Agree, it is annoying, to achieve that I need to implement custom double click listener, It is not difficult, trying to experiment with that

evidolob:
@lstocchi @jeffmaury 
I made all changes, PR should be ready for review.

evidolob:
@lstocchi I did what you suggested

PR #125, opened 2020-06-08
added watch implementation to update tree automatically (#73)

### LINKED ISSUE #73: Refresh tekton tree automatically
The Tekton Pipelines tree currently requires a manual refresh to display the latest state of the pipelines and pipelineruns. This would leave the user thinking the pipeilnerun is running for a long time.

lstocchi:
it resolves #73 

I had to use the latest version of fabric8 tekton as the previous ones had some issues with betav1 resources. Currently it doesn't support triggers.

I tried this patch by launching pipelines, deleting resources externally from ide and the tree refreshes quite good. Let me know what you think.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ intellij {
 }
 
 dependencies {
-    compile 'io.fabric8:openshift-client:4.9.0'
+    compile 'io.fabric8:openshift-client:4.10.2'
     compile 'io.fabric8:tekton-client:4.10.2'
     compile 'com.redhat.devtools.intellij:intellij-common:0.0.1-SNAPSHOT'
 }
--- a/build.gradle
+++ b/build.gradle
@@ -27,6 +27,7 @@ intellij {
 
 dependencies {
     compile 'io.fabric8:openshift-client:4.9.0'
+    compile 'io.fabric8:tekton-client:4.10.2'
     compile 'com.redhat.devtools.intellij:intellij-common:0.0.1-SNAPSHOT'
 }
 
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -27,6 +27,7 @@ intellij {
 
 dependencies {
     compile 'io.fabric8:openshift-client:4.9.0'
+    compile 'io.fabric8:tekton-client:4.10.2'
```
jeffmaury, line 30: Please make sure we use a consistent version between OpenShift client and Tekton client

jeffmaury:
I tested and it work fine for me but  I deleted a task and expected that corresponding TaskRun(s) will disappear as well but it was not the case but it seems to be on the Tekton Pipelines side because `tkn taskrun list` still reports them

lstocchi:
> I tested and it work fine for me but I deleted a task and expected that corresponding TaskRun(s) will disappear as well but it was not the case but it seems to be on the Tekton Pipelines side because `tkn taskrun list` still reports them

jeffmaury:
> 
> 
> > I tested and it work fine for me but I deleted a task and expected that corresponding TaskRun(s) will disappear as well but it was not the case but it seems to be on the Tekton Pipelines side because `tkn taskrun list` still reports them
> 
> I didn't think about it. We should report it to them? Idk if there is some reason why they don't delete the taskruns too.
> 
> Thinking about it, it is not so easy because the task could also be referenced by some pipeline and then the taskrun by a pipelinerun, So, how should you proceed? Deleting everything?

We should ask them as the taskrun resources have annotations generated by Tekton Pipelines that establish the relation with the object that created it but regarding the pipeline it's a little bit complex

PR #130, opened 2020-06-11
Open PipelineRun and TaskRun as readonly

evidolob:
This PR add ability to open PipelineRun and TaskRun resource in editor with readonly mode.
It supports both way to open(from context menu and from double click).

Fix: #82

Demo:
ezgif com-video-to-gif (22)


evidolob:
@lstocchi @jeffmaury 
I have doub about TaskRun for Condition, to be able to open it in editor(to get YAML) I need to have full task run name, previously we use `conditionName` and Name to that node, in this PR I change it, I pass as node name task run name(something like `conditional-pipeline-run-pvtdf-then-check-m24sp-file-exis-sbb6q`) and use `conditionName` as `stepName` in `TaskRun` constructor. That works, but may be a bit confusing. 
Another option may be to have new node, data class for Condition TaskRun.
WDYT?

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/TreeHelper.java
```diff
@@ -96,9 +99,14 @@ public static Tree getTree(Project project) {
...
             kind = KIND_TASKRUN;
+            readonly = true;
+        } else if (node instanceof PipelineRunNode){
+            content = tkncli.getPipelineRunYAML(namespace, node.getName());
+            kind = KIND_PIPELINERUN;
+            readonly = true;
         }
 
-        return Pair.pair(content, kind);
+        return new Triple<>(content, kind, readonly);
```
lstocchi, line 109: It is difficult for me to comment here (all the code related to Pair/Triple) as I'm not a seasoned java developer so maybe @jeffmaury can shed some light. Coming from c# where i used `structs` and javascript with `objects` this makes completely sense to me (i was the one who used Pair at first i think) but i'm not sure if in java this is considered good code or not as it can be confusing to know which values (first, second, third) represent. I have the fear that by keeping adding stuff to this, the triple could become a four-value tuple in future, then five-values and so on... @jeffmaury WDYT? How should one behave in such situation?

Looking at the code, i would have kept it as it was and just check the kind inside `createAndOpenVirtualFile` and if it was a `taskrun` or `pipelinerun` set the writable to false. This way you could have changed less code i guess. 

BTW don't keep this comment in consideration as this is just my written thought to understand better what to do in such cases. Better to wait the review by Jeff

lstocchi:
> @lstocchi @jeffmaury
> I have doub about TaskRun for Condition, to be able to open it in editor(to get YAML) I need to have full task run name, previously we use `conditionName` and Name to that node, in this PR I change it, I pass as node name task run name(something like `conditional-pipeline-run-pvtdf-then-check-m24sp-file-exis-sbb6q`) and use `conditionName` as `stepName` in `TaskRun` constructor. That works, but may be a bit confusing.
> Another option may be to have new node and data class for Condition TaskRun.
> WDYT?

For me we can keep it as you did. Looks fine to me.

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/TreeHelper.java
evidolob: Unfortunately java do not have  `structs` or easy way to return new 'object' class. 
I'm also to use TS a lot, and for me it is strange to create a new class, just for return multiple values from function/method. In java common way to solve this, it use dedicated class for result. And I think if I create an use class witch will describe Tekton resource which we want to open will simplify `createAndOpenVirtualFile`, we may reduce number of params in it.

> For `Pair` and `Triple` I was looking on some presentation from JetBrains guy, and he discuss why they have this class in IT platform, he say that it is comfort way to return multiple value from function/method, without introducing additional type, which will be used as container for values. But if you want to return more that 3 values than you need to have own type. But, of course, it is matter of code guidelines that we decide to apply on this case.

jeffmaury:
@evidolob or @lstocchi can you rebase ?

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/VirtualFileHelper.java
```diff
@@ -35,14 +38,24 @@ public static void openVirtualFileInEditor(Project project, String namespace, St
...
-            VirtualFile vf = createTempFile(name, content);
+            VirtualFile vf;
+
+            //open TaskRun and PipelineRun in read only mode
+            if (KIND_PIPELINERUN.equals(kind) || KIND_TASKRUN.equals(kind)) {
+                vf = new LightVirtualFile(name, content);
+                vf.setWritable(false);
+            } else {
+                vf = createTempFile(name, content);
+                File fileToDelete = new File(vf.getPath());
```
lstocchi, line 51: These two rows can be removed. To fix an issue with the cache we moved the deleteOnExit() inside the `createTempFile` method https://github.com/redhat-developer/intellij-tekton/blob/a52327b63b3131b87077e53f837b12d4b246d1c8/src/main/java/com/redhat/devtools/intellij/tektoncd/utils/VirtualFileHelper.java#L59

evidolob: Done

PR #131, opened 2020-06-15
Add preview editor for pipelines

### LINKED ISSUE #76: Add pipeline diagram preview
Pipeline YAML could be long and complex which makes it difficult to get an overview of all tasks included in the pipeline and their execution order. It would be beneficial to be able to get a read only visual graph of the pipeline based on the YAML to help user rapidly understand the tasks and how they'd execute.

jeffmaury:
Fixes #76

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/ui/editors/PipelineGraphUpdater.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/ui/editors/PipelineGraphUpdater.java
@@ -110,7 +110,7 @@ public class PipelineGraphUpdater implements GraphUpdater<Pipeline> {
                     vertex = graph.insertVertex(null, node.id, node.name, 0, 200, WIDTH, HEIGHT, "editable=false;");
                     break;
                 case CONDITION:
-                    vertex = graph.insertVertex(null, node.id, node.name, 0, 200, WIDTH, HEIGHT, "shape=rhombus;");
+                    vertex = graph.insertVertex(null, node.id, node.name, 0, 200, WIDTH, HEIGHT, "shape=rhombus;editable=false;");
                     break;
             }
             vertexes.put(node.id, vertex);
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/ui/editors/PipelineGraphUpdater.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/ui/editors/PipelineGraphUpdater.java
@@ -16,6 +16,7 @@ import com.mxgraph.view.mxGraph;
 import io.fabric8.tekton.pipeline.v1beta1.Pipeline;
 import io.fabric8.tekton.pipeline.v1beta1.PipelineTask;
 import io.fabric8.tekton.pipeline.v1beta1.PipelineTaskCondition;
+import io.fabric8.tekton.pipeline.v1beta1.PipelineTaskInputResource;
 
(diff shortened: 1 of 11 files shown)


evidolob:
A few thing that I noticed:

- when I move task rectangle, it is move the rectangle itself, but keeps resize squares on old place:
Screenshot 2020-06-15 at 10 51 28
- It is possible to double click on task rectangle and preview allows to edit task name, but doesn't reflect in yaml source. I think we should disable that, or reflect changes to yaml.
-The graph it self doesn't correctly build with conditions:
Screenshot 2020-06-15 at 10 50 44
In this case `file-exists` should be run after `first-create-file`, as `file-exists` use result of `first-create-file`. See [guard-task-execution-using-conditions](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#guard-task-execution-using-conditions) for details. 
- When I click on task rectangle and immediately start dragging it create arrow, and I can attach it to any node, or just keep it un attached:
Screenshot 2020-06-15 at 11 03 18


> Pipeline sample: https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/conditional-pipelinerun.yaml

lstocchi:
Not sure if we have the same issue here but i'm linking this issue so we can verify before merging https://github.com/redhat-developer/vscode-tekton/issues/361

jeffmaury:
Should be fixed by #132 and #148

src/main/java/com/redhat/devtools/intellij/tektoncd/ui/editors/PipelineGraphUpdater.java
```diff
@@ -0,0 +1,167 @@
...
+        try {
+            return MAPPER.readValue(content, Pipeline.class);
+        } catch (IOException e) {
+            return null;
+        }
+    }
+
+    @Override
+    public void update(Pipeline content, mxGraph graph) {
+        List<PipelineTask> tasks = content.getSpec().getTasks();
```
lstocchi, line 67: Here we should check if spec section exists otherwise we can have an error. 
If you create a new pipeline, delete the spec section and start by changing the name you have 
image

lstocchi (changes requested):
Tested. Overall it looks good. 
There are still some minor stuff that doesn't convince me, same to the things Yevhen noticed. 
- when I move task rectangle, the rectangle itself is moved but the resize squares are still displayed in the old place (same when resizing)
- you can move the link between two squares and connect two different squares (It could make a user think he can change the configuration by working on the graph)

BTW for me we can also merge this as it is (there is only a bug below) and work on the improvements later. 
The advantages are larger than the "issues".

src/main/java/com/redhat/devtools/intellij/tektoncd/ui/editors/PipelineGraphUpdater.java
jeffmaury: Can't reproduce the move block issue, I tested both on Windows and Linux but noticed that when you move a condition, the connector is not correctly placed. See 
tekton1
Will provide a fix for the empty spec case.

lstocchi: I tried it again and incredibly i am not able to replicate it, even by changing theme, now it works .... don't know why but that's good lol

lstocchi (approved):
LGTM

Issue #136, opened 2020-06-23
When file is pushed to the cluster, IJ shows warning that the file has been modified outside


jeffmaury:

lstocchi:
Closing as https://github.com/redhat-developer/intellij-tekton/pull/139 was merged

PR #139, opened 2020-06-25
Fixed issue with cache when saving file (#136)

lstocchi:
Long story short.

Based on https://www.jetbrains.org/intellij/sdk/docs/basics/virtual_file_system.html, the virtual file system encapsulates most of the activity related to files. One of the main goal is "Tracking file modifications and providing both old and new versions of the file content when a modification is detected" which is what we have in the current implementation.

Because the vfs manages a persistent snapshot of the user's hard disk what we were missing was to delete the file from the tmp folder and clean (refresh) the vfs so to delete the cached version.


robstryker:
I've copied this fix into my repo :D   Thanks!

PR #148, opened 2020-07-10
Regenerate schemas based on pipleline 0.13.2 and triggers 0.5.0

jeffmaury:
- Added a test to validate the schemas
- Schemas are now prettyfied so that diff is relevant


lstocchi:
Before to review this... is there any reason why we don't go directly with pipeline 0.14.1 and triggers 0.6.1? I think they introduced the finally clause in the new release

jeffmaury:
Yes because this PR is quite old and it changes the format of the generated schemas files to be pretty printed (not anymore on a single line) so that when we change the Pipeline or Triggers version we have a better view of the changes (diff will be readable). So this one will be followed by other or not.

PR #165, opened 2020-07-28
Added code completion for taskRef (#153)

### LINKED ISSUE #153: taskRef code completion should list tasks in current cluster

lstocchi:
it fixes #153

### DIFF
--- a/src/it/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCliPipelineTest.java
+++ b/src/it/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCliPipelineTest.java
@@ -13,6 +13,7 @@ package com.redhat.devtools.intellij.tektoncd.tkn;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.junit.Test;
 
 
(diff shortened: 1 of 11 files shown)


jeffmaury:
With an empty pipeline, I have this result:
tekton1

lstocchi:
Fixed some issue related to when make tasks list visible. Now they are shown after the name tag (must be child of a taskRef node). Below how it works for me with your task. I wasn't able to replicate your behavior. Maybe i fixed it somehow with my previous edits?? 
 
taskrefcodecompl1

jeffmaury:
If the taskRef already has some params, upon code completion a new params section is inserted and it causes syntax validation.
tekton1

lstocchi:
> If the taskRef already has some params, upon code completion a new params section is inserted and it causes syntax validation.

Not sure how to handle that and if we should delete something from the editor. Maybe the user edited something or did that on purpose? If there are two equal sections is because it was his choice to add the same task twice, as in your examples. The current behavior works exactly as in IJ.

code-comple-ij

One thing i noticed is that i only show tasks, I think we should add clustertasks as well, right?

jeffmaury:
TBH I would favor a more fine grained code completion / syntax validation:

1. on taskRef/name: update only the name field based on the list of tasks available
2. add syntax validation based on the selected task and checking if required parameters / resources are specified
3. on parameters/resources names, perform the same kind of code completion as 1

Asking for comments @evidolob @siamaksade

lstocchi:
I don't know... to me it over complicates the code for no advantages. I really don't see your reasoning for this. Users are going to write params in any case and in your way they need to write a lot of yaml by hand. 

It also make me think users will need to open the task itself to give it a look and check all the params/resources required to start writing the pipeline, which slow down the process. It's like when you implement an interface and IntelliJ copy all methods signatures to be implemented. I prefer to delete something extra than writing everything from scratch.
Let's see what Yevhen and Siamak think

jeffmaury (approved):
LGTM now

PR #170, opened 2020-08-13
Update schemas to process timeout as string

### LINKED ISSUE #169: timeout property is generated as object should be string syntax

jeffmaury:
Fixes #169


jeffmaury:
Good point but according to the Go Duration doc, the syntax is the following;

```
// ParseDuration parses a duration string.
// A duration string is a possibly signed sequence of
// decimal numbers, each with optional fraction and a unit suffix,
// such as "300ms", "-1.5h" or "2h45m".
// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
```

Code uses the following regexp:
`// [-+]?([0-9]*(\.[0-9]*)?[a-z]+)+`

See https://github.com/golang/go/blob/master/src/time/format.go#L1374

lstocchi:
I saw that but it didn't make much sense in this plugin for me. Who is going to use negative numbers? Or "ns", "us" (or "µs"), "ms" time units? But i guess we can add the default pattern and then see if some user complains before to think a different regex pattern

src/main/resources/schemas/tekton.dev/v1alpha1_Pipeline.json
```diff
@@ -1258,8 +1253,8 @@
           "type": "array"
         },
         "timeout": {
-          "$schema": "http://json-schema.org/draft-04/schema#",
-          "$ref": "#/definitions/Duration"
+          "pattern": "^[-+]?([0-9]*(\\.[0-9]*)?(ns|us|µs|μs|ms|s|m|h))$",
```
lstocchi, line 1256: Not sure which is your idea on it but if we want to follow Go Duration doc it doesn't work correctly. For instance it fails with 1h3m

src/main/resources/schemas/tekton.dev/v1alpha1_Pipeline.json
```diff
@@ -1253,7 +1253,7 @@
           "type": "array"
         },
         "timeout": {
-          "pattern": "^[-+]?([0-9]*(\\.[0-9]*)?(ns|us|µs|μs|ms|s|m|h))$",
+          "pattern": "^[-+]?([0-9]*(\\.[0-9]*)?(ns|us|µs|μs|ms|s|m|h))+$",
```
lstocchi, line 1256: My only doubt are the first asterisks. In the current form, `1hm2s` is valid or also `1.h`, maybe `^[-+]?([0-9]+(\\.[0-9]+)?(ns|us|µs|μs|ms|s|m|h))+$` is better??

Other "issue" i noticed (which i write so we are aware of it) is the order of the units - 10m1h is accepted. But i guess we should trust users for this.

jeffmaury: According to the Go code, both are valid:

- .1s or 1.s are valid
- no ordering for units

PR #176, opened 2020-08-17
Switch to cli 0.11.0 (#168)

### LINKED ISSUE #168: Update cli to 0.11.0

lstocchi:
it resolves #168


jeffmaury:
Would be good to extend the integration suite each time we upversion the CLI as this is only way to validate the CLI does not bring some issues by being not compatible. Can you add a couple of integration tests ?

lstocchi:
> Would be good to extend the integration suite each time we upversion the CLI as this is only way to validate the CLI does not bring some issues by being not compatible. Can you add a couple of integration tests ?

Added some. Maybe we can create some issue to extend them and spend some time on it during the sprint.

jeffmaury (approved):

PR #177, opened 2020-08-18
Fix issue when config change during runtime

### LINKED ISSUE #144: Error with fabric8 when logging in cluster happens after the plugin started
I just noticed we can have an issue with fabric8 when the login is made after the plugin started.

1. Start the cluster (without log in)
2. Start the plugin -> it will notice the user is not logged in and the error node will be shown correctly
3. Log in to the cluster
4. The tree is correctly refreshed and all nodes are visible (pipelines, tasks ...)
5. If you try to expand a node it will prompt an error -> Fabric8 not authorized or some error related to watchers.

lstocchi:
It resolves #144 
This is no a direct patch for #144 as i guess it was solved partially with #145 but i found another issue while testing it. 

I found out that if you changed the config on runtime the tree didn't refresh correctly. The problem is that the client (DefaultKubernetesClient) is initialized with a config but if it changes somehow, the tkn client doesn't get recreated. So if you are logged in as an admin and you change to a developer, the tkn client (initialized as admin) keeps working with that config.

To test it, i started crc without being logged in. Then i started the plugin. The tree contained only the root and a warning node to tell i wasn't logged in. Then i logged in as admin and then switched to developer. 
Before this patch the tree didn't refresh. With this, the tree nodes are displayed correctly.

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/tree/TektonRootNode.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/tree/TektonRootNode.java
@@ -25,6 +25,7 @@ public class TektonRootNode {
   }
 
   public CompletableFuture<Tkn> initializeTkn() {
+    TknCliFactory.getInstance().setTkn(project);
     return TknCliFactory.getInstance().getTkn(project).whenComplete((tkn, err) -> this.tkn = tkn);
   }
 
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/tree/TektonRootNode.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/tree/TektonRootNode.java
@@ -25,7 +25,6 @@ public class TektonRootNode {
   }
 
   public CompletableFuture<Tkn> initializeTkn() {
-    TknCliFactory.getInstance().setTkn(project);
     return TknCliFactory.getInstance().getTkn(project).whenComplete((tkn, err) -> this.tkn = tkn);
   }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/redhat/devtools/intellij/tektoncd/tree/TektonRootNode.java
```diff
@@ -25,6 +25,7 @@ public TektonRootNode(Project project) {
   }
 
   public CompletableFuture<Tkn> initializeTkn() {
+    TknCliFactory.getInstance().setTkn(project);
```
jeffmaury, line 28: I don't get why do you need to insert your code as calling getTkn will cause setTkn to be called

lstocchi: setTkn (in getTkn) is only called if future is null. If you change your config once it has been already initialized, future will be different from null and setTkn will never be called. This raise a possible issue.

An example.... when you start the plugin, tkn client is initialized by using the current config (suppose admin/cluster), if you at a certain time switch user to `developer` (current config becomes developer/cluster), the client is never updated and it will remain the same initialized with the old config which is wrong.

jeffmaury: I think it would be more understandable to do:
1) add a reset method that would be called on the load method of the root node and that method would reset the future to null
2) keep initializeTkn as it was

WDYT ?

lstocchi: Ok for me

jeffmaury (approved):
Tested while switching from not logged to admin

PR #180, opened 2020-08-24
added runAfter code completion (#152)

### LINKED ISSUE #152: RunAfter code completion should list tasks in the current pipeline

lstocchi:
it resolves #152 

This patch adds code completion for the runAfter tag. When the code completion is requested, it gets all tasks placed earlier in the pipeline than the one selected and shows them in the options. Below a demo.

runAfter


jeffmaury:
Are you really looking only at previous tasks, I'm not sure the semantic of the `runAfter` field is holding this restriction

lstocchi:
Good point, it really don't care. I'm going to change it

jeffmaury (approved):
LGTM. However, I think it would be interesting to add unit tests

PR #186, opened 2020-08-31
add option to delete all related resources of a pipeline/task (#171)

### LINKED ISSUE #171: Allow users to delete all related resources when deleting a pipeline/task
We should allow users to delete all resources associated to a pipeline when deleting it.

`--prs                           Whether to delete pipeline(s) and related resources (pipelineruns) (default: false)`

The same with tasks

`--trs                           Whether to delete Task(s) and related resources (TaskRuns) (default: false)`

lstocchi:
it resolves #171 

Below some screenshot

Deleting a single task
image

Deleting a Pipeline
image

Deleting some resources
image

Deleting a lot of resources
image


jeffmaury (approved):
LGTM but would be good to have one integration test checking the effect of the new flag

jeffmaury (approved):

PR #187, opened 2020-09-02
add support for variable substitution code completion (#154)

### LINKED ISSUE #154: Add code completion to list all params/resources/workspaces/results in pipeline/task steps

lstocchi:
it resolves #154 

It adds code completion for variables in pipeline, task and condition. Examples below. 

Even though they are not 100% accurate (e.g they don't mention conditions at all) I took the rules from https://github.com/tektoncd/pipeline/blob/master/docs/variables.md

Pipeline 
pipeline-var

Task
task-var

Condition
condition-var

It looks huge but 80% of the file edited are new yamls used for tests.


jeffmaury (approved):
Nice new feature. Tested on pipeline and task

PR #192, opened 2020-09-07
Open start wizard with data from previous *run

### LINKED ISSUE #161: Allow opening start pipeline wizard with params from previous pipelineruns
Currently it's possible to rerun a previous pipelinerun but it does not give a chance to the user to modify a subset of the parameters, etc from the previous pipelinerun before the pipelinerun starts executing. It would be valuable to enable opening the start pipeline wizard with a previous pipelinerun params, etc to allow user modify the values before start.

This could possibly be an action on the pipelinerun context menu named "Restart PipelineRun" or similar.

lstocchi:
it resolves #161 

This PR allows users to open the start wizard with data taken from an old *run in two ways

1) A new action belonging to a *run node. It opens up the wizard with data taken from the *run selected. 
If the original pipeline has been updated with new fields, only the ones existing in the *run will be filled in.
If the pipeline has been updated and some fields have been removed, only the ones existing in the pipeline/*run will be filled in.

I called the action `start mirror` but i am not really sure about the naming. Any better idea?
image

2) The wizard has a new panel (advanced options) which contains an option to take data from a specific *run.
Once a *run is selected in the combobox, the wizard is updated with the values taken from that *run.

The options panel is quite empty now but all options related to the start action will be added here.

image

image

image

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/Constants.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/Constants.java
@@ -31,6 +31,8 @@ public class Constants {
 
     public static final String KIND_PIPELINE = "pipeline";
     public static final String KIND_PIPELINERUN = "pipelinerun";
+    public static final String KIND_TASK = "task";
+    public static final String KIND_CLUSTERTASK = "clustertask";
     public static final String KIND_TASKRUN = "taskrun";
 
     public static final String KIND_PVC =  "PersistentVolumeClaim";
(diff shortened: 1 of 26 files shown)


jeffmaury:
My remarks as discussed:

- if the pipeline/task does not requires input (params/workspaces/resources), then it should be launched directly
- if the pipeline/task does not have runs, when you use the Start action, you can access the import feature where it should be hidden

jeffmaury:
Also tested it on a pipeline but new run is never launched

jeffmaury (changes requested):
No start launched

jeffmaury (changes requested):
Regarding the case where a pipeline does not have existing runs, I would hide the option to import rather than keep the option and having an error message when you enable this option

jeffmaury (approved):
LGTM

PR #200, opened 2020-09-10
switched back to 0.11.0

lstocchi:
Due to a performance issue with tkn cli 0.12.0, this patch switch the cli used by the plugin back to 0.11.0.

By testing again with the old cli, you can really see the difference in speed.


jeffmaury (commented):
LGTM

jeffmaury (approved):

PR #203, opened 2020-09-10
change order behavior based on taskrun node position (#202)

### LINKED ISSUE #202: Taskruns in reverse order below TaskRuns node
Nodes below the taskruns node are in reversed order.  This is a consequence of https://github.com/redhat-developer/intellij-tekton/pull/185 . We should have different behavior based on the taskrun node position.
image

lstocchi:
it resolves #202 

Now the taskrun nodes ordering depends on the position of the node.
Below the TASKRUNS node -> from newest to oldest
Below a TASK node -> from newest to oldest
Below a PIPELINERUN node -> from oldest to newest (so one can check which was the first task executed in the pipeline)


jeffmaury (approved):
LGTM

PR #215, opened 2020-09-17
add support for finally tasks in wizard (#204)

### LINKED ISSUE #204: Start pipeline wizard does not support finally tasks
Finally tasks for a pipeline are not available through the start wizard::

Pipeline:

```
kind: Pipeline
metadata:
  name: app-deploy
spec:
  resources:
  - name: pipeline-resource-type
    type: git
  tasks:
  - name: foo
    taskRef:
      kind: Task
      name: build-app
  finally:
    - name: cleanup
      taskRef:
        kind: Task
        name: echo

```

Wizard page:

image

lstocchi:
it resolves #204

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/utils/StartResourceModel.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/utils/StartResourceModel.java
@@ -104,12 +104,24 @@ public class StartResourceModel extends ConfigurationModel{
 
     private void initTaskServiceAccounts(String configuration) throws IOException {
         if (isPipeline()) {
+            this.taskServiceAccountNames = new HashMap<>();
             JsonNode tasksNode = YAMLHelper.getValueFromYAML(configuration, new String[] {"spec", "tasks"});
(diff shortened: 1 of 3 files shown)


jeffmaury (approved):
LGTM. The only issue I had (same for the graphical representation) is that the order of the tasks in the Authentication tab is not the same as the order of declaration.
Another issue that I found (but it's not related to this PR) is that a pipeline without inputs does not display the wizard so we can't specify the authentication info if we need)

lstocchi:
> LGTM. The only issue I had (same for the graphical representation) is that the order of the tasks in the Authentication tab is not the same as the order of declaration.
> Another issue that I found (but it's not related to this PR) is that a pipeline without inputs does not display the wizard so we can't specify the authentication info if we need)

+1 for the first

The second issue is something i thought when I was creating the wizard but i always skipped it because i couldn't think of an actual  use case where someone could need authorization without any input. But i agree it should work as you described. One can always add hardcoded values (git repo url, ...) to a pipeline/task. I'll open a new issue.

PR #216, opened 2020-09-17
stop showing generic snippets with other code completion (#191)

### LINKED ISSUE #191: Remove general snippets from other code completion handlers
Right now the general snippets are always displayed even when they are not really useful and other code completion handlers are called. We should change this behavior and show them only if they are really needed based on the position or the file (e.g show empty pipeline/task/resources snippet only on an empty file...)

lstocchi:
it resolves #191


jeffmaury (approved):
Nicer completion now

Issue #213, opened 2020-09-16
Allow to configure tkn location/find it in PATH


lstocchi:
The plugin only checks $HOME/.tekton location and if nothing is in there, it downloads the cli needed. 
IJ should be smart enough to try to find it in the default location or in PATH. If nothing is found it should ask the user to set a location or download it.

PR #218, opened 2020-09-18
fixed regex to extract tkn version (#213)

lstocchi:
This is the first part related to #213. It should find tkn if present in PATH


jeffmaury (approved):
Tested on Windows and it fixes the case when tkn 0.11.0 is in the PATH

PR #223, opened 2020-09-18
Add code completion for when clause

### LINKED ISSUE #209: Add code completion for when clause
Pipeline added a new field `when` for simpler conditional execution of tasks. (See https://github.com/tektoncd/pipeline/pull/3135)
We may provide some code completion, note sure because the doc mentions that both Input and Values are static.

lstocchi:
it resolves #209 

@jeffmaury this patch was born to add code completion for when clause but it is starting going too far so i prefer to stop here and complete the initial work in further PRs.

This patch contains two things essentially.
1) the actual code completion for when clause -> this shows the `in`, `notin` operator options
2) Additional features useful for the when clause:
   - code completion for tasks results (often used as input/value in a when clause)
   - implementation of the virtual file system to avoid downloading a task multiple times when unnecessary

Once this will be merged i will extend the usage of the vfs to everything else and add tests for the vfs.

An example of how it works below:

whenoperator


jeffmaury (changes requested):
According to the source code, the values for Operator are the following:
```

	!
	=
	==
	in
	!=
	notin
	exists
	gt
	lt
```

But the doc mentions only `in` and `notin`.
So asking @jerop for confirmation

jerop:
> According to the source code, the values for Operator are the following:
> 
> ```
> 
> 	!
> 	=
> 	==
> 	in
> 	!=
> 	notin
> 	exists
> 	gt
> 	lt
> ```
> 
> But the doc mentions only `in` and `notin`.
> So asking @jerop for confirmation

@jeffmaury even though `Operator` has all those options, we only support `in` and `notin` in `When Expressions` given the use cases we looked at for guarding `Task` execution

If another `Operator` is provided, the `Pipeline` validation would fail and throw an error

We could consider supporting other `Operators` as needed, based on the use cases, in the future

jerop:
@lstocchi please note that `input`, `operator` and `values` are supposed to be lowercase in the yaml -- it seems in the implementation and example of the completion that they're pascalcase 🤔

jeffmaury:
Yes that a side effect because our JSON schema generator is parsing the Go code and I noticed that the `WhenExpression` type does not have the json annotation like other types in the pipeline code. @jerop let me know if I should open an issue

lstocchi:
> Yes that a side effect because our JSON schema generator is parsing the Go code and I noticed that the `WhenExpression` type does not have the json annotation like other types in the pipeline code. @jerop let me know if I should open an issue

I didn't pay much attention on this aspect before but now that @jerop pointed that out i guess we need to fix it. If someone copy/paste an example from the tekton repo (which contains the lowercase names), she/he will get a warning in our plugin. Not good.

jerop:
> Yes that a side effect because our JSON schema generator is parsing the Go code and I noticed that the `WhenExpression` type does not have the json annotation like other types in the pipeline code. @jerop let me know if I should open an issue

@jeffmaury added the annotations in https://github.com/tektoncd/pipeline/pull/3291

excited to see code completion for when expressions supported in the plugin!

jeffmaury (approved):

PR #224, opened 2020-09-21
Action to take infos regarding resources created by a run

### LINKED ISSUE #181: Action to gather logs/info/errors about resources created by a run
Not sure if it should be added here but it would be really useful if users had a way to find out why their pipelines/tasks are stuck. It happened to me earlier. I started a pipeline and after 15m it was still running. By going to the cluster and make some calls i found out that the pod wasn't created by an error about the affinity rule. In the statefulset I found out the reason... the label made by tekton was too long. Error below.

It made me think. By using google and knowing the cluster you can find out the reason, eventually, but we should try to make developers life easier. Couldn't we add an action that get automatically infos regarding the pod/statefulset created by a run? 

```
Pod affinity-assistant-local-maven-repo-petclinic-deploy-dev-run-4bk42-0 in StatefulSet affinity-assistant-local-maven-repo-petclinic-deploy-dev-run-4bk42 failed error: Pod "affinity-assistant-local-maven-repo-petclinic-deploy-dev-run-4bk42-0" is invalid: [metadata.labels: Invalid value: "affinity-assistant-local-maven-repo-petclinic-deploy-dev-run-4bk42-0": must be no more than 63 characters, metadata.labels: Invalid value: "affinity-assistant-local-maven-repo-petclinic-deploy-dev-run-4bk42-695f4d675f": must be no more than 63 characters, spec.hostname: Invalid value: "affinity-assistant-local-maven-repo-petclinic-deploy-dev-run-4bk42-0": must be no more than 63 characters]
  Warning  FailedCreate  10s (x15 over 2m54s)  statefulset-controller  create Pod affinity-assistant-local-maven-repo-petclinic-deploy-dev-run-4bk42-0 in StatefulSet affinity-assistant-local-maven-repo-petclinic-deploy-dev-run-4bk42 failed error: Pod "affinity-assistant-local-maven-repo-petclinic-deploy-dev-run-4bk42-0" is invalid: [metadata.labels: Invalid value: "affinity-assistant-local-maven-repo-petclinic-deploy-dev-run-4bk42-695f4d675f": must be no more than 63 characters, metadata.labels: Invalid value: "affinity-assistant-local-maven-repo-petclinic-deploy-dev-run-4bk42-0": must be no more than 63 characters, spec.hostname: Invalid value: "affinity-assistant-local-maven-repo-petclinic-deploy-dev-run-4bk42-0": must be no more than 63 characters]
```

lstocchi:
It resolves #181 

This is the a very basic implementation of an action which should be useful to debug a *run. 
As an example, i created a pvc with a size which wasn't supported by my cluster (there was no pv available for that size) and so the *run could not be completed but it was stuck in a running mode. 
By using the tkn cli the user doesn't receive any feedback in such cases.
By clicking on the `Show Diagnostic Data` action this is the result. 
image

I'll try to add more examples as soon as i find them. Any other suggestion for this? @jeffmaury

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/ShowDiagnosticDataAction.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/ShowDiagnosticDataAction.java
@@ -1,0 +1,40 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ * Red Hat, Inc.
+ ******************************************************************************/
+package com.redhat.devtools.intellij.tektoncd.actions;
+
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.redhat.devtools.intellij.common.utils.ExecHelper;
+import com.redhat.devtools.intellij.tektoncd.tkn.Tkn;
+import com.redhat.devtools.intellij.tektoncd.tree.ParentableNode;
+import com.redhat.devtools.intellij.tektoncd.tree.PipelineRunNode;
+import com.redhat.devtools.intellij.tektoncd.tree.TaskRunNode;
+import javax.swing.tree.TreePath;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ShowDiagnosticDataAction extends TektonAction {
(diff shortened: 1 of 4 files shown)


jeffmaury (changes requested):
I had a couple of issues:

- it turns out that for a reason I ignore, I have pipelineruns in my cluster with no associated pods or statefulsets thus the generated string is empty. We should not open a terminal in this case but rather report to the user that no data was extracted.
- launching echo on Windows does not work because it's not a process so I think the correct way to do it is to launch cmd.exe /C echo xxxx

lstocchi:
Changed it based on your comment. Now a warning popup is displayed if no data is found.
For the windows command i don't have idea if it works, just let me know.

lstocchi:
@jeffmaury is this ready?

jeffmaury:
Yes but needs a rebase.

PR #243, opened 2020-10-08
fix visibility issue cancel action

lstocchi:
This is something i noticed yesterday. The cancel action wasn't visible when it should have. Fixed it.


jeffmaury (approved):
LGTM

PR #246, opened 2020-10-12
Update diagram when a task depends on the result of another

### LINKED ISSUE #221: Update diagram when a task depends on the result of another
If in a pipeline a task depends on the result of another task (e.g `$(tasks.taskname.results.somevalue)`) the diagram doesn't show this dependency (it should look like a runAfter) 

An example below. The task `print-res` depend on `numbers-equal` but the diagram doesn't show it.

image

jeffmaury:
Fixes #221


lstocchi (approved):
Good catch. I didn't think at this use case. LGTM

Issue #227, opened 2020-09-23
Add support for clusterTask for taskRef Code completion


lstocchi:

lstocchi:
Closing it as #248 was merged

PR #248, opened 2020-10-12
add support for workspaces when using taskRef completion

### LINKED ISSUE #225: Add workspaces field when displaying task by using taskref code completion

lstocchi:
it resolves #225 and #227 

Now the taskRef code completion also shows clusterTasks (if present). To distinguish them from normal tasks i added a label ot the right.

This patch also add support for workspaces.

clustertasktaskref

About tests, I didn't find a way to inject my mocked classes/methods inside the call made by the CodeInsightTestFixture, so making this PR ready for review without tests right now. I'll keep looking at it, sooner or later i will discover the lacking piece.


jeffmaury:
Tested the code completion on **taskRef.name**:
- worked as I can now see the cluster tasks being listed.
- but the issue I have is that the kind **attribute** was specified to **Task** and is now duplicated, is it fixable ?

lstocchi:
> Tested the code completion on **taskRef.name**:
> 
> * worked as I can now see the cluster tasks being listed.
> * but the issue I have is that the kind **attribute** was specified to **Task** and is now duplicated, is it fixable ?

Yes, i think we can check if the `kind` has already been added before and replace it but not sure if it's worth the effort as it will be uncomplete. This use case happens if you are editing an existing task in a pipeline. What if the old task also had some params/resources/workspaces? Are we going to replace them too? Or only the `kind`?

jeffmaury:
If we are to follow the replace pattern then we must do it for every field. But I think this can be part of another issue.
Can you rebase it so that we can process the release ?

lstocchi:
> If we are to follow the replace pattern then we must do it for every field. But I think this can be part of another issue.
> Can you rebase it so that we can process the release ?

Rebased :+1:

jeffmaury (approved):

PR #267, opened 2020-10-30
Grey unused variables

### LINKED ISSUE #182: Check and show if input/output are really used inside the configuration file
While writing the configuration of a pipeline/task it would be useful to highlight somehow the input/outputs that are not used in the configuration. Similar to what happen when writing code and a variable is greyed because it is never used.

lstocchi:
it resolves #182 

This patch makes all unused variables being colored in gray so users can easily see if they have some variables that can be removed or that has to be used somewhere. Useful when you copy a pipeline/task from internet/TektonHub and start editing it based on your requirement.

unusedvariables

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/component/field/Input.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/component/field/Input.java
@@ -63,6 +63,10 @@ public class Input {
     }
 
     public Input fromJson(JsonNode inputNode, Kind kind) {
+        if (!inputNode.has("name") ||
+                inputNode.get("name").isNull()) {
+            return null;
+        }
         String name = inputNode.get("name").asText();
         String type = "string";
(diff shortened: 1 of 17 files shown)


jeffmaury (changes requested):
Failed for me for pipeline.
Here is my pipeline:

```
kind: Pipeline
spec:
  params:
    - name: p1
  tasks:
  - name: error
    taskRef:
      kind: Task
      name: error
    params:
      - name: p1
        value: ${params.p1}
```

lstocchi:
> apiVersion: tekton.dev/v1beta1
> kind: Pipeline
> spec:
>   params:
>     - name: p1
>   tasks:
>   - name: error
>     taskRef:
>       kind: Task
>       name: error
>     params:
>       - name: p1
>         value: ${params.p1}

The syntax is wrong in this pipeline. It misses the name and you are using curly brackets instead of parenthesis, it must be $(...

jeffmaury (approved):
LGTM

PR #269, opened 2020-11-03
Add generated CRDs for better compatibility with IntelliJ Kubernetes …

### LINKED ISSUE #263: cannot use intellij-tekton to edit Pipelines if intellij-kubernetes is installed
to be able to open the `intellij-tekton` editor you have to disable the [intellij-kubernetes plugin](https://plugins.jetbrains.com/plugin/10485-kubernetes) which isn't awesome; as that plugin is super handy for general purpose k8s and editing helm charts etc.

I wonder if https://github.com/redhat-developer/intellij-tekton/issues/250 fixes this? If not maybe there needs to be a Tekton file type so that we can explicitly open pipelines in `intellij-tekton` editor without having to uninstall `intellij-kubernetes` (as it opens all k8s yaml files by default right now)

jeffmaury:
…plugin

- K8S metadata specified as object thus annotations and labels are reported as errors
- Some time related fields are not propertly handled
- tkn cli removes the version info for Triggers resources

Fixes #263

**WARNING**

In order to test it, you need to change **ideaVersion** in _gradle.properties_ from IC-XXXX to IU-XXXX as the IntelliJ Kubernetes plugin is compatible with IU only

### DIFF
--- a/.github/workflows/IJ.yml
+++ b/.github/workflows/IJ.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        IJ: [IC-2019.2, IC-2019.3, IC-2020.1, IC-2020.2]
+        IJ: [IU-2019.2, IU-2019.3, IU-2020.1, IU-2020.2]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


lstocchi:
Definetly an improvement, now the code completion works fine. The only problem is that all values are reported as error. Do you have the same behavior? Can this be solved?
image

lstocchi (approved):
Tested. LGTM

PR #270, opened 2020-11-05
Prepare for 0.3.0

jeffmaury:


lstocchi (approved):
LGTM.

Something that we need to do (just noticed it, we can also do it after the release) is to update the commands section in the README. We always forgot to update it in the past releases as well https://github.com/redhat-developer/intellij-tekton#actions-available-for-a-tekton-pipelinetaskclustertaskresource , (e.g it misses `Get Diagnostic data`, `Add trigger`, ...)

PR #283, opened 2020-11-30
Integrate TektonHub

### LINKED ISSUE #121: Tekton catalog integration
Tekton catalog provides a catalog of reusable tasks and pipelines. It would be good to integration it when user creates a pipelines.

lstocchi:
it resolves #121

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/WindowToolFactory.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/WindowToolFactory.java
@@ -30,6 +30,7 @@ import com.redhat.devtools.intellij.tektoncd.listener.TreeDoubleClickListener;
 import com.redhat.devtools.intellij.tektoncd.listener.TreeExpansionListener;
 import com.redhat.devtools.intellij.tektoncd.listener.TreePopupMenuListener;
 import com.redhat.devtools.intellij.tektoncd.tree.TektonTreeStructure;
+import com.redhat.devtools.intellij.tektoncd.ui.hub.HubModel;
 import org.jetbrains.annotations.NotNull;
 
 import java.lang.reflect.Constructor;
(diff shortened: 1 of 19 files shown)


lstocchi:
@jeffmaury this is how it works. Still some issue to be solved and code to be refactored (i am sure there is duplicate code to be removed somewhere) but if you can give it a look before tomorrow lunch so to know if there is something to change (visually) before the demo
tektonhub

src/test/java/com/redhat/devtools/intellij/tektoncd/utils/model/ConditionConfigurationModelTest.java
```diff
@@ -23,6 +29,14 @@
 
     @Test
     public void checkEmptyConditionModel() throws IOException {
+        ResourceApi resApi = new ResourceApi();
```
jeffmaury, line 32: Seems to me this is useless

src/main/java/com/redhat/devtools/intellij/tektoncd/WindowToolFactory.java
```diff
@@ -48,6 +49,7 @@ public void createToolWindowContent(@NotNull Project project, @NotNull ToolWindo
             tree.putClientProperty(Constants.STRUCTURE_PROPERTY, structure);
             tree.setCellRenderer(new NodeRenderer());
             tree.addTreeWillExpandListener(new TreeExpansionListener());
+            HubModel.getInstance();
```
jeffmaury, line 52: What is the rationale ? Is model is loaded asynchronously then `getInstance` should return a `CompletableFuture` and should called when required

jeffmaury (approved):
See my comments inline.
Other remarks:
- when you install a task and repeat install of the same workflow, then warning message is not show
- when a task is installed, I don't see the notification
- we don't support installation as cluster tasks but that may be postponed to another issue
- in terms of UI, if we can shrink a little but: I think the Install buttons are a little but too large and the area for displaying a task in the list also

lstocchi:
> See my comments inline.
> Other remarks:
> 
> * when you install a task and repeat install of the same workflow, then warning message is not show
> * when a task is installed, I don't see the notification
> * we don't support installation as cluster tasks but that may be postponed to another issue
> * in terms of UI, if we can shrink a little but: I think the Install buttons are a little but too large and the area for displaying a task in the list also

I think I covered everything but the clusterTask installation. Please give it a look. Thanks!!

jeffmaury:
Looks better but still:
- can't see the notification
- I think UI can still be shrinked a little bit

image

lstocchi:
> Looks better but still:
> 
> * can't see the notification

What do you mean by notification? A popup window in the right bottom corner? Or you don't see the notification message in the event log panel?

jeffmaury:
> 
> 
> > Looks better but still:
> > 
> > * can't see the notification
> 
> What do you mean by notification? A popup window in the right bottom corner? Or you don't see the notification message in the event log panel?

The popup window

lstocchi:
@jeffmaury resized the column a little bit more. I don't think i can go further otherwise the warning icon is hidden.
Also added the popup notification. Give it a look, thanks!!

jeffmaury (approved):
Notification is now displayed

PR #293, opened 2020-12-09
Refresh node before expanding (#275)

### LINKED ISSUE #275: Running pipeline is not shown under pipeline
(I have followed this turorial for creating pipeline: https://github.com/openshift/pipelines-tutorial#deploy-sample-application)
I have created pipeline:
```
kind: Pipeline
metadata:
  creationTimestamp: "2020-11-06T09:50:07Z"
  generation: 1
  managedFields:
  - apiVersion: tekton.dev/v1beta1
    fieldsType: FieldsV1
    fieldsV1:
      f:spec:
        .: {}
        f:params: {}
        f:tasks: {}
        f:workspaces: {}
    manager: okhttp
    operation: Update
    time: "2020-11-06T09:50:07Z"
  name: app-build
  namespace: pipelines-tutorial
  resourceVersion: "39247839"
  selfLink: /apis/tekton.dev/v1beta1/namespaces/pipelines-tutorial/pipelines/app-build
  uid: 0806b110-b3e3-44fc-a1f9-17a39a80d3e2
spec:
  params:
  - description: url of the git repo for the code of deployment
    name: git-url
    type: string
  - default: master
    description: revision to be used from repo of the code for deployment
    name: git-revision
    type: string
  - description: image to be build from the code
    name: IMAGE
    type: string
  tasks:
  - name: fetch-repository
    params:
    - name: url
      value: $(params.git-url)
    - name: subdirectory
      value: ""
    - name: deleteExisting
      value: "true"
    - name: revision
      value: $(params.git-revision)
    taskRef:
      kind: ClusterTask
      name: git-clone
    workspaces:
    - name: output
      workspace: shared-workspace
  - name: build-image
    params:
    - name: TLSVERIFY
      value: "false"
    - name: IMAGE
      value: $(params.IMAGE)
    runAfter:
    - fetch-repository
    taskRef:
      kind: ClusterTask
      name: buildah
    workspaces:
    - name: source
      workspace: shared-workspace
  workspaces:
  - name: shared-workspace

```

When I have tried to run it, the running pipeline shows in PipelineRuns tree, but not under Pipelines->app-build (this pipeline), but app-build shows under Pipelines->app-build when its finished.

lstocchi:
it resolves #275 

The problem here is that the watch is set when a node is actually expanded but its children were already calculated and the user see an old version of them.

E.g in a tree like
```
root
|->namespace
          |-> pipelines
                     |-> pipeline-1
                     |-> pipeline-2
```
when the tree is created it contains `root`, `namespace` and `pipelines` nodes, with `pipelines` node hidden. When `namespace` is expanded, `pipelines` is made visible and `pipeline-1` and `pipeline-2` are pre-loaded (they are hidden). At this point if something change on cluster (e.g `pipeline-3` is created), the children of pipelines are not updated as this was not expanded and the watch has not been activated.

This is a quick fix. As soon as we start depending on the k8s plugin and we show only the current active namespace we can switch at having the watch set up at node creation and keep it always alive

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/listener/TreeExpansionListener.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/listener/TreeExpansionListener.java
@@ -10,8 +10,11 @@
  ******************************************************************************/
 package com.redhat.devtools.intellij.tektoncd.listener;
 
+import com.intellij.ui.treeStructure.Tree;
 import com.redhat.devtools.intellij.common.actions.StructureTreeAction;
+import com.redhat.devtools.intellij.tektoncd.Constants;
 import com.redhat.devtools.intellij.tektoncd.tree.ParentableNode;
+import com.redhat.devtools.intellij.tektoncd.tree.TektonTreeStructure;
 import com.redhat.devtools.intellij.tektoncd.utils.WatchHandler;
 
 import javax.swing.event.TreeExpansionEvent;
@@ -23,6 +26,7 @@ public class TreeExpansionListener implements TreeWillExpandListener {
         ParentableNode<? extends ParentableNode<?>> expandingElement = StructureTreeAction.getElement(treeExpansionEvent.getPath().getLastPathComponent());
         if (WatchHandler.get().canBeWatched(expandingElement)) {
             WatchHandler.get().setWatchByNode(expandingElement, treeExpansionEvent.getPath());
+            ((TektonTreeStructure)((Tree)treeExpansionEvent.getSource()).getClientProperty(Constants.STRUCTURE_PROPERTY)).fireModified(expandingElement);
         }
     }
 


src/main/java/com/redhat/devtools/intellij/tektoncd/listener/TreeExpansionListener.java
```diff
@@ -23,6 +26,7 @@ public void treeWillExpand(TreeExpansionEvent treeExpansionEvent) {
         ParentableNode<? extends ParentableNode<?>> expandingElement = StructureTreeAction.getElement(treeExpansionEvent.getPath().getLastPathComponent());
         if (WatchHandler.get().canBeWatched(expandingElement)) {
             WatchHandler.get().setWatchByNode(expandingElement, treeExpansionEvent.getPath());
+            ((TektonTreeStructure)((Tree)treeExpansionEvent.getSource()).getClientProperty(Constants.STRUCTURE_PROPERTY)).fireModified(expandingElement);
```
jeffmaury, line 29: I don't like this patch at all because it seems to me that refresh will be cause each time an expand is requested, defeating the notion of cache. And it links two events that should not (expand and state (refresh).

jeffmaury (approved):

PR #305, opened 2020-12-17
Generate taskrun example from task (#289)

### LINKED ISSUE #289: Generate an example TaskRun YAML for a Task
As a Tekton user, I want to select a Task and generate an example TaskRun YAML for that particular task so that I can modify and tweak it based on my needs before applying it to the cluster.

See https://github.com/redhat-developer/vscode-tekton/issues/354

lstocchi:
it resolves #289 

As discussed, this patch shows the *run that is going to be created in the preview panel of the start dialog

image

To be in sync with vscode I also added the action to generate a taskrun template so the user can edit it in the editor. 

image

Then once the action is clicked a skeleton of the taskRun is opened and it works exactly as the other resources. When the user is ready to save, the IDE asks if it has to be pushed to the cluster.

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/Constants.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/Constants.java
@@ -43,10 +43,10 @@ public class Constants {
     public static final String KIND_CLUSTERTRIGGERBINDING = "clustertriggerbinding";
     public static final String KIND_EVENTLISTENER = "eventlistener";
 
-    public static final String KIND_PVC =  "PersistentVolumeClaim";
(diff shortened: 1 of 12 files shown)


jeffmaury (changes requested):
Tested on task aws-cli from Hub but found issues with workspaces:
EmptyDir: "{}"

Should be:
emptyDir: {}

jeffmaury (changes requested):
Still have `EmptyDir` instead of `emptyDir`:

image

lstocchi:
> Still have `EmptyDir` instead of `emptyDir`:

I completely missed the name was wrong. Fixed for all workspace kinds

jeffmaury (approved):

PR #312, opened 2021-01-04
use kubernetes-client 5.0.0 (#311)

### LINKED ISSUE #311: use (latest) kubernetes-client 5.0
[kubernetes-client 5.0.0](https://search.maven.org/artifact/io.fabric8/kubernetes-client/5.0.0/jar) was released on 20201230.

adietish:
depends on https://github.com/redhat-developer/intellij-common/issues/24

fixes #311

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,8 +27,8 @@ intellij {
 }
 
 dependencies {
-    compile 'io.fabric8:openshift-client:4.12.0'
-    compile 'io.fabric8:tekton-client:4.12.0'
+    compile 'io.fabric8:openshift-client:5.0.0'
+    compile 'io.fabric8:tekton-client:5.0.0'
     compile 'com.redhat.devtools.intellij:intellij-common:0.0.1-SNAPSHOT'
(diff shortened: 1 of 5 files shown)


adietish:
The test errors point to changes in the kubernetes-client. I'll first split `AbstractPipelineGraphUpdater#generateTree` into separate methods so that I better grasp this rather complicated code. I'll then proceed to fix the error.

jeffmaury:
You also need to check/update intellij-common as it also use kubernetes-client as a dependency

adietish:
@lstocchi, @jeffmaury the test error was caused by improved deserialization in the newer library:
`when` properties are now available via a getter whereas they were only accessible in the additional properties where any unknown property is stored in.
Dare to test this thoroughly please?

lstocchi (approved):
Played with it with both CRC and minikube just to be sure. Everything is working as expected. I think i tried all the features we offer in this plugin. We also have all the following sprint to test it along the way before the next release so ... LGTM.

PR #319, opened 2021-01-12
fix: Allows parameter array to accept empty array as value (#313)

### LINKED ISSUE #313: Parameter with default is requested
The **curl** task from the Tekton Hub has an **options** parameter which has a default value.
However, when if want to start this task, the wizard wants me to fill a value

lstocchi:
it resolves #313 

I am quite unsure the empty array in yaml is converted correctly but i copied what tekton does. Basically if you pass an empty array to the array param, the pipeline/task will be run like this

```
- name: values
    value:
    - ""
```
which for me is not an empty array but an array with an empty string as first value. Maybe something to ask to the tekton team.

Other thing to think about is if we want to allow users to start pipelines/tasks with empty string parameters.
With this patch only array parameter accepts an empty value, in a string param the empty value is considered an error.
WDYT? @jeffmaury If we consider that there could be a default value as an empty string it makes sense to remove this check

### DIFF
--- a/src/it/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCliPipelineTest.java
+++ b/src/it/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCliPipelineTest.java
@@ -11,6 +11,7 @@
 package com.redhat.devtools.intellij.tektoncd.tkn;
 
 import com.redhat.devtools.intellij.tektoncd.TestUtils;
+import com.redhat.devtools.intellij.tektoncd.tkn.component.field.Input;
 import io.fabric8.tekton.client.TektonClient;
 import java.io.IOException;
 import java.util.Arrays;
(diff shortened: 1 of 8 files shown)


jeffmaury:
Not sure who is converting an empty array to an array of a single empty string which is not semantically equivalent.
YAML supports empty arrays through []

lstocchi:
maybe the cli, if you try to run the `curl` task, it tells the default value is an empty string. I'll investigate more.
image

What about `string` type params? Do we want to accept empty strings in the wizard or we show an error if the textboxes are left empty?

lstocchi:
FYI opened an issue at https://github.com/tektoncd/cli/issues/1286

lstocchi:
It requires the new tkn cli 0.18.0 to work. https://github.com/redhat-developer/intellij-tekton/pull/424
Now the empty array is used correctly
image

lstocchi:
@jeffmaury rebased, it is ready to be reviewed

jeffmaury:
Still can't start the task without entering some value in the options field

PR #320, opened 2021-01-12
fix pipeline preview when value is too long (#314)

### LINKED ISSUE #314: Pipeline preview sometimes is not usable
Show Preview is sometime not usable: here is an example on the curl task from the TektonHub:

image

lstocchi:
it resolves #314 

Now values too long will be written as multilines.
image


jeffmaury (approved):
LGTM

PR #321, opened 2021-01-13
fix UI glitch in add trigger wizard (#239)

### LINKED ISSUE #239: UI glitch in the Add trigger wizard
I noticed a small UI glitch in the Add trigger wizard when a TriggerBinding is about to be created and the first time the Save link is clicked:

tekton

This has been reproduced both on Windows10Pro and RHEL8. 
Both are using IC-2019.2 and Java 11.0.3+12-b304.10

lstocchi:
it resolves #239


jeffmaury (approved):
LGTM

PR #322, opened 2021-01-13
add basic preferences (#247)

### LINKED ISSUE #247: Add preferences for the plugin
I think we should improve the user experience by adding preferences for the plugin.
Here is my initial contribution
- default behavior when deleting resources (related resources) so that user does not have to select the checkbox (default false)
- always show start wizard: in the case of a task/pipeline without required input, we still show the wizard in order to configure authentication information. We could relax that. Default true

lstocchi:
it resolves #247

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/StartAction.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/StartAction.java
@@ -94,10 +94,10 @@ public class StartAction extends TektonAction {
                                     && model.getInputResources().isEmpty()
                                     && model.getOutputResources().isEmpty()
                                     && model.getWorkspaces().isEmpty();
-            boolean showWizardWithNoInputs = SettingsState.getInstance().showStartWizardWithNoInputs || !hasNoInputs;
+            boolean showWizard = SettingsState.getInstance().showStartWizardWithNoInputs || !hasNoInputs;
 
             StartWizard startWizard = null;
-            if (showWizardWithNoInputs) {
+            if (showWizard) {
                 startWizard = UIHelper.executeInUI(() -> {
                     String titleDialog;
                     if (element instanceof PipelineNode) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/redhat/devtools/intellij/tektoncd/actions/StartAction.java
```diff
@@ -89,19 +90,28 @@ public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object s
...
-                }
-                StartWizard wizard = new StartWizard(titleDialog, element, getEventProject(anActionEvent), model);
-                wizard.show();
-                return wizard;
-            });
+            boolean hasNoInputs = model.getParams().isEmpty()
+                                    && model.getInputResources().isEmpty()
+                                    && model.getOutputResources().isEmpty()
+                                    && model.getWorkspaces().isEmpty();
+            boolean showWizardWithNoInputs = SettingsState.getInstance().showStartWizardWithNoInputs || !hasNoInputs;
```
jeffmaury, line 97: Rename to showWizard for better readability

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/WatchHandler.java
```diff
@@ -226,6 +224,7 @@ public void eventReceived(Action action, T resource) {
                 // watches for *runs are always active so there also could be nothing to refresh, only a notification to display
                 if (!SettingsState.getInstance().displayPipelineRunResultAsNotification ||
                         !(resource instanceof io.fabric8.tekton.pipeline.v1beta1.PipelineRun) ||
+                        ((PipelineRun) resource).getStatus() == null ||
```
jeffmaury, line 227: Not sure to understand this change

jeffmaury (changes requested):

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/WatchHandler.java
lstocchi: The getStatus can return null somehow and if we don't check it,the next getCompletionTime call fails with a nullpointerex. Not sure why I didn't have it earlier.

jeffmaury (approved):

PR #327, opened 2021-01-20
add code completion on single input (#193)

### LINKED ISSUE #193: Add code completion for single inputs (params, workspaces, resources)
This extends https://github.com/redhat-developer/intellij-tekton/issues/153 
It can be useful to have this code completion when working on an old configuration where tasks used have been updated and the user needs to add only few new inputs

lstocchi:
it resolves #193 

This patch adds code completion on single inputs. So if in your pipeline you have a task that has been updated you can just add the newly inputs without writing everything from scratch.
The code completion also add the default value if available.

single-input


jeffmaury:
I have small issues with this PR. First I took curl task from Hub and tried to build a pipeline.

tekton

lstocchi:
rebased and fixed your issue :crossed_fingers:

jeffmaury (approved):
LGTM.
Still have the issue that name is not proposed but I'm not sure this is linked to this PR

PR #335, opened 2021-02-08
chore: Add workflow to create changelog and plugin archive when new tag is created

lstocchi:
Output sample https://github.com/lstocchi/intellij-tekton/releases/tag/0.0.81


.github/workflows/tag.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ steps.get_version.outputs.VERSION }}
+          release_name: ${{ steps.get_version.outputs.VERSION }}
+          body: ${{ steps.changelog.outputs.changelog }}
+      - name: Attach vsix to release
```
jeffmaury, line 41: Should be Attach zip to release

jeffmaury (changes requested):
Also I tried it on my repo but got an error: see https://github.com/jeffmaury/intellij-tekton/runs/1962419062?check_suite_focus=true

lstocchi:
> Also I tried it on my repo but got an error: see https://github.com/jeffmaury/intellij-tekton/runs/1962419062?check_suite_focus=true

I think it fails bc there is only 1 tag. The action should take the last tag + the previous and take all commits between them. I guess it misses the logic for when there is only 1 tag. I'll look into it but if you try by creating a new tag it should work.

jeffmaury:
Adding a new tag but still failing: https://github.com/jeffmaury/intellij-tekton/runs/1963486856?check_suite_focus=true

So giving up and reviewing based on your sample. LGTM but Features paragraph should be first, not after Bug Fixes ?

lstocchi:
Checked the code. In the function to retrieve all valid tags, it filters by checking if the version is semver valid
`const releaseTags = tags.filter((tag) => compareVersions.validate(tag.name));`

the validate does this 
```
/**
   * Validate [semver](https://semver.org/) version strings.
   * 
   * @param version Version number to validate
   * @returns `true` if the version number is a valid semver version number, `false` otherwise.
   *
   * @example
   * ```
   * compareVersions.validate('1.0.0-rc.1'); // return true
   * compareVersions.validate('1.0-rc.1'); // return false
   * compareVersions.validate('foo'); // return false
   * ```
   */
```

as you are using invalid semver version it fails.

jeffmaury:
OK got it as I was using vx.x as tag names !!!!

lstocchi:
The order of the sections depend on the commits history right now. The fix was first bc the first commit retrieved from gh was a fix. I'll change that by using the order found in the workflow when we specify the types used. So if we have [feat, fix, docs] the sections will be showed in this order.

lstocchi:
Update action with new section ordering. 
Example here https://github.com/lstocchi/intellij-tekton/releases/tag/0.0.82

jeffmaury (approved):
LGTM. Maybe rename as release.yml, looks more natural for novice users landing on the project

lstocchi:
@jeffmaury Ok, should i also add a section to the documentation that explain this?

jeffmaury:
I think the only thing to document is the conventional commits so should go into CONTRIBUTING.md

PR #336, opened 2021-02-08
add workflow to check conventional commit syntax is used

lstocchi:


.github/workflows/conventionalCheck.yml
```diff
@@ -0,0 +1,16 @@
+name: "Conventional Commits PR Check"
+
+on:
+  pull_request_target:
```
jeffmaury, line 4: Never use pull_request_target as it's a security hole. Contact @sbouchet for details.

jeffmaury (changes requested):

.github/workflows/conventionalCheck.yml
lstocchi: I think you are referring to the red warning [here](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target), right? If that's the problem we should be ok as we don't checkout/build/do anything. We only check the PR/commit message.
I'm using the pull_request_target because, in this way, it won't use the configuration from some fork, what's in the master will be always used.

jeffmaury: The problem is that Red Hat security team is scanning GitHub and sent us an email as soon as @sbouchet pushed a PR like this one. And what if any user sends a PR that modify this workflow and retrieve secrets, damage the repo,... ?

lstocchi: This is why i used it. Based on what i've understood, the workflow that will be used will always be the one from the master even if the PR has a different version. So no-one can change the conventional types/scopes we decide to use.

sbouchet: @lstocchi @jeffmaury: you're both right :)
security team scan github repos and look how the pull_request target are used in workflow, especially with secrets.
but in that case, as also pointed [https://securitylab.github.com/research/github-actions-preventing-pwn-requests](here) (and it's the link security is referring too), it's clearly indicated that:
**Combining pull_request_target workflow trigger with an explicit checkout of an untrusted PR is a dangerous practice that may lead to repository compromise.**
so i think we're safe because we're only using the action here.

jeffmaury (approved):

.github/workflows/conventionalCheck.yml
lstocchi: Thanks for the clarification @sbouchet :+1:

Issue #280, opened 2020-11-16
Display active namespace in tree view.


sudhirverma:
It would be good to display the active namespace in the tree view to have similar behavior as in vscode-tekton. Rather than displaying all namespace in the tree view.

lstocchi:
+1. This is something we already thought about. The problem we have is that we don't have any k8s plugin to rely on to allow users to switch namespace/project from the ide itself. As soon as andre's k8s plugin will be ready we can use that one to accomplish this.

PR #337, opened 2021-02-09
fix: add tektonHub label to new installed task/clustertask (#330)

### LINKED ISSUE #330: Use Hub provided label to select tasks/pipelines installed from Tekton Hub

lstocchi:
it resolves #330 

The tekton cli only works with tasks right now so i needed to edit/save the clustertask yaml "manually". 

We also need to agree on the conventional commits types because I used `feat` in this case but i'm not sure it's the correct one. This is not a real feature. It's more something that's useful for us for building future stuff.
I found this descriptions

```
feat: Add a new feature to the codebase (MINOR in semantic versioning).
fix: Fix a bug (equivalent to a PATCH in Semantic Versioning).
docs: Documentation changes.
style: Code style change (semicolon, indentation...).
refactor: Refactor code without changing public API.
perf: Update code performances.
test: Add test to an existing feature.
chore: Update something without impacting the user (ex: bump a dependency in package.json).
```
So in this case, this should have been named `chore`? WDYT? @jeffmaury

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/Constants.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/Constants.java
@@ -61,6 +61,8 @@ public class Constants {
 
     public static final String STRUCTURE_PROPERTY = Constants.class.getPackage().getName() + ".structure";
 
+    public static final String HUB_CATALOG_TAG = "hub.tekton.dev/catalog";
+
     public enum InstallStatus {
         ERROR,
         INSTALLED,
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/TektonHubAction.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/TektonHubAction.java
@@ -28,6 +28,9 @@ import javax.swing.tree.TreePath;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+
+import static com.redhat.devtools.intellij.tektoncd.Constants.HUB_CATALOG_TAG;
+
 public class TektonHubAction extends TektonAction {
     Logger logger = LoggerFactory.getLogger(TektonHubAction.class);
 
@@ -40,8 +43,14 @@ public class TektonHubAction extends TektonAction {
             String namespace = element.getNamespace();
             List<String> tasks, clusterTasks;
             try {
-                tasks = tkncli.getTasks(namespace).stream().map(task -> task.getMetadata().getName()).collect(Collectors.toList());
-                clusterTasks = tkncli.getClusterTasks().stream().map(ct -> ct.getMetadata().getName()).collect(Collectors.toList());
+                tasks = tkncli.getTasks(namespace).stream()
+                        .filter(task -> task.getMetadata().getLabels() != null && task.getMetadata().getLabels().containsKey(HUB_CATALOG_TAG))
+                        .map(task -> task.getMetadata().getName())
+                        .collect(Collectors.toList());
+                clusterTasks = tkncli.getClusterTasks().stream()
+                        .filter(ct -> ct.getMetadata().getLabels() != null && ct.getMetadata().getLabels().containsKey(HUB_CATALOG_TAG))
+                        .map(ct -> ct.getMetadata().getName())
+                        .collect(Collectors.toList());
             } catch (IOException e) {
                 UIHelper.executeInUI(() ->
                         Messages.showErrorDialog(
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/TektonHubAction.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/TektonHubAction.java
@@ -52,7 +52,7 @@ public class TektonHubAction extends TektonAction {
             }
 
             Project project = getEventProject(anActionEvent);
-            HubModel model = new HubModel(project, namespace, tasks, clusterTasks, element instanceof TasksNode);
+            HubModel model = new HubModel(project, tkncli, namespace, tasks, clusterTasks, element instanceof TasksNode);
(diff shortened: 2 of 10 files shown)


jeffmaury (changes requested):
I tried the following scenario:
- ClusterTasks nodes was empty
- install a task from Hub as cluster task
Did not get the ClusterTasks node refreshed (still empty)

Also see some hardcoding of the catalog name to tekton which seems to be wrong to me as the catalog name is returned by the Hub API

I'm also wondering if this new label doesn't affect our algorithm that detect if the task is already installed ?

lstocchi:
> I tried the following scenario:
> 
> * ClusterTasks nodes was empty
> * install a task from Hub as cluster task
>   Did not get the ClusterTasks node refreshed (still empty)
> 
> Also see some hardcoding of the catalog name to tekton which seems to be wrong to me as the catalog name is returned by the Hub API
> 
> I'm also wondering if this new label doesn't affect our algorithm that detect if the task is already installed ?

1. If you refresh the ct node you see the hub's clusterTask installed correctly? Is it just a problem with the refresh?
2. yeah, correct. I did that voluntarily as i hope i can remove everything as soon as the cli supports ct. The doubt is, who is responsible to set up another catalog? Isn't the user who should have access to a setting where he can specify his own catalog so the plugin can point to that? We don't have this feature yet. I need to check.
3. No that shouldn't be a problem, right now we just check the name. We can use the label to improve it. So if there is another task with the same name that wasn't installed by the Hub we can offer a different behavior.

jeffmaury:
> 
> 
> > I tried the following scenario:
> > 
> > * ClusterTasks nodes was empty
> > * install a task from Hub as cluster task
> >   Did not get the ClusterTasks node refreshed (still empty)
> 
> > 
> 
>     1. If you refresh the ct node you see the hub's clusterTask installed correctly? Is it just a problem with the refresh?
Yes only after refresh the task is visible

> 
> 
> > Also see some hardcoding of the catalog name to tekton which seems to be wrong to me as the catalog name is returned by the Hub API
> 
>     1. yeah, correct. I did that voluntarily as i hope i can remove everything as soon as the cli supports ct. The doubt is, who is responsible to set up another catalog? Isn't the user who should have access to a setting where he can specify his own catalog so the plugin can point to that? We don't have this feature yet. I need to check.
No I think a single Hub endpoint is allowed to host several catalogs

> 
> 
> > I'm also wondering if this new label doesn't affect our algorithm that detect if the task is already installed ?
> 
>     1. No that shouldn't be a problem, right now we just check the name. We can use the label to improve it. So if there is another task with the same name that wasn't installed by the Hub we can offer a different behavior.
That something we need to align with VSCode so @evidolob may help

lstocchi:
> Yes only after refresh the task is visible

Ok, it is always the same problem with the watchers that are enabled when the node is expanded. This will be fixed with https://github.com/redhat-developer/intellij-tekton/issues/280 . When i'll refactor the tree i'll change the watchers to be always active. We can keep this open until we have that one merged.

> No I think a single Hub endpoint is allowed to host several catalogs

I have doubts on this point. I mean, i didn't find anything (just a mention about 2  different catalogs alphav1 and betav1 and only the second is active on the hub) .... but if what you say can be true my question is, how can i decide which catalog to use to take a task? I suppose that two catalogs can have tasks with the same name. So what if i want to download one on a specific catalog? :man_shrugging: I don't find any docs regarding this. Do you know who can answer this or we do a general question on the tekton channel?

> That something we need to align with VSCode so @evidolob may help

:+1:

jeffmaury:
> 
> 
> > Yes only after refresh the task is visible
> 
> Ok, it is always the same problem with the watchers that are enabled when the node is expanded. This will be fixed with #280 . When i'll refactor the tree i'll change the watchers to be always active. We can keep this open until we have that one merged.
But my node was expanded because I had a single cluster task that I deleted then go to the hub to create a new one

> 
> > No I think a single Hub endpoint is allowed to host several catalogs
> 
> I have doubts on this point. I mean, i didn't find anything (just a mention about 2 different catalogs alphav1 and betav1 and only the second is active on the hub) .... but if what you say can be true my question is, how can i decide which catalog to use to take a task? I suppose that two catalogs can have tasks with the same name. So what if i want to download one on a specific catalog? 🤷‍♂️ I don't find any docs regarding this. Do you know who can answer this or we do a general question on the tekton channel?
When you query the Hub I think you can specify the catalog to look into and when tasks are returned from Hub, they store a reference to the catalog they are from
> 
> > That something we need to align with VSCode so @evidolob may help
> 
> 👍

evidolob:
@lstocchi @jeffmaury 
Not sure that I understand your problem. If you are using `tkn hub install task` command to install task it will add all labels.
In vscode-tekton we manually add `hub.tekton.dev/catalog` label in case if we install task as cluster task. The value of it(catalog) you can take from `ResourceData.catalog` object.

jeffmaury:
> 
> 
> @lstocchi @jeffmaury
> Not sure that I understand your problem. If you are using `tkn hub install task` command to install task it will add all labels.
> In vscode-tekton we manually add `hub.tekton.dev/catalog` label in case if we install task as cluster task. The value of it(catalog) you can take from `ResourceData.catalog` object.

@evidolob my question is once you've installed a task do you use this label do detect a Hub task is already installed in your cluster. Today our algorithm is using names to detect

evidolob:
Yes we use that label, to detect that that task installed from hub, but we check only label presents, and do not check the value

lstocchi:
@jeffmaury the problem with the clustertask node should be solved with the new tree. For the other problems, it now took the catalog from the tekton item installed and it filters the task/ct by using the label (as in vscode) to check if the task/ct is already installed.

jeffmaury (approved):
LGTM. I would choose fix; because it fixes the algorithm used to match tasks installed from Hub

Issue #331, opened 2021-02-02
Should use telemetry


adietish:
In https://github.com/adietish/intellij-redhat-telemetry a telemetry plugin is being developed. We should use it in intellij-tekton.

lstocchi:
Are we adding telemetry to IJ Tekton now? Don't we need to have an "OK" from some PMs? From what i understood in last cabal, Fred said only few plugins were allowed to use telemetry at the beginning for testing, no? Is IJ Tekton one of those?

@jeffmaury Do you already have in mind which things/actions do we need to catch/log?

jeffmaury:
No this is OK now. I would start recording actions first

lstocchi:
@jeffmaury i thought about it when you were talking about recording IJ versions. It could be useful for us to check how many ppl uses this plugin on tekton with triggers enabled and which versions of them they are using (the same with tekton pipelines version) ... as there were breaking changes in past, so we know what to do and how urgent it is (e.g https://github.com/redhat-developer/intellij-tekton/issues/252)

lstocchi:
Closing it as https://github.com/redhat-developer/intellij-tekton/pull/341 was merged. Thanks @adietish :rocket:

PR #341, opened 2021-02-10
feat: notify telemetry of all actions (#331)

adietish:
depends on https://github.com/redhat-developer/intellij-redhat-telemetry/pull/2/commits which has to be merged before this can compile.


adietish:
now sending the following events:
* "open editor" when an editor is opened:
```
 "properties": {
    "app_name": "IntelliJ IDEA",
    "app_version": "2019.3",
    "duration": "00:00:00",
    "extension_name": "Tekton Pipelines by Red Hat",
    "extension_version": "0.7.0-SNAPSHOT",
    "java_version": "11.0.4",
    "resource_kind": "clustertasks",
    "resource_namespace": "default"
  },
```
* "save to cluster" when an editor is saved and the result saved to the cluster:
```
  "properties": {
    "app_name": "IntelliJ IDEA",
    "app_version": "2019.3",
    "duration": "00:00:00",
    "extension_name": "Tekton Pipelines by Red Hat",
    "extension_version": "0.7.0-SNAPSHOT",
    "java_version": "11.0.4",
    "new_resource": "new",
    "resource_kind": "ClusterTask",
    "resource_version": "tekton.dev/v1beta1",
    "result": "success"
  },
```
* "delete resource" when a resource is deleted:
```
"properties": {
    "app_name": "IntelliJ IDEA",
    "app_version": "2019.3",
    "duration": "00:04:23",
    "extension_name": "Tekton Pipelines by Red Hat",
    "extension_version": "0.7.0-SNAPSHOT",
    "java_version": "11.0.4",
    "result": "ClusterTaskNode resources deleted."
  },
```

@jeffmaury, @lstocchi : please have a look at the properties, do they make sense?

adietish:
here's the error that's reported when saving an editor with a name that already exists. I guess that we should also remove namespace and resource name:
```
  "properties": {
    "app_name": "IntelliJ IDEA",
    "app_version": "2019.3",
    "duration": "00:00:00",
    "error": "An error occurred while saving ClusterTask foo22\nclustertasks.tekton.dev \"foo22\" already exists\n",
    "extension_name": "Tekton Pipelines by Red Hat",
    "extension_version": "0.7.0-SNAPSHOT",
    "java_version": "11.0.4",
    "resource_kind": "ClusterTask",
    "resource_version": "tekton.dev/v1beta1"
  },
```

lstocchi:
> now sending the following events:
> 
> * "open editor" when an editor is opened:
> 
> ```
>  "properties": {
>     "app_name": "IntelliJ IDEA",
>     "app_version": "2019.3",
>     "duration": "00:00:00",
>     "extension_name": "Tekton Pipelines by Red Hat",
>     "extension_version": "0.7.0-SNAPSHOT",
>     "java_version": "11.0.4",
>     "resource_kind": "clustertasks",
>     "resource_namespace": "default"
>   },
> ```
> 
> 
> @jeffmaury, @lstocchi : please have a look at the properties, do they make sense?

@jeffmaury about the open in editor action, would it make sense to have a field that say why the editor was opened? Like, editor opened to create new resource or opened to edit a task ?

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/DeleteAction.java
```diff
@@ -106,45 +112,53 @@ public void actionPerformed(AnActionEvent anActionEvent, TreePath[] path, Object
...
-                        } else if (type.equals(TriggerBindingNode.class)) {
-                            tkncli.deleteTriggerBindings(namespace, resources);
-                        } else if (type.equals(ClusterTriggerBindingNode.class)) {
-                            tkncli.deleteClusterTriggerBindings(resources);
-                        } else if (type.equals(EventListenerNode.class)) {
-                            tkncli.deleteEventListeners(namespace, resources);
-                        }
+                        deleteResources(type, resourcesByClass, namespace, deleteRelatedResources, tkncli);
                         ((TektonTreeStructure) getTree(anActionEvent).getClientProperty(Constants.STRUCTURE_PROPERTY)).fireModified(resourcesByClass.get(type).get(0).getParent());
+                        telemetry.property(TelemetryService.PROP_RESOURCE_KIND, type.getSimpleName())
```
lstocchi, line 124: it misses if the user also wants to delete relatable resources.

lstocchi (changes requested):

lstocchi:
The other thing i think we should improve is how the message to enable telemetry works. If you deny it, the message is not shown anymore. Even in a new session. We should atleast add a checkbox with "remember my answer". If the checkbox is checked and the answer is deny, i agree with not showing it anymore but it may happen that the user click deny bc he is in rush and doesn't want to read the privacy in that exact moment but he may be interested in it another day. 
Other thing that worries me is the location of the setting to enable/disable it. If i'm using the tekton plugin, i would expect to find the setting there, not under Red Hat telemetry. It's not very intuitive. I could have tens of plugins and i could be not aware that the tekton/k8s/openshift is developed by redhat. We should add some info in the plugin setting.

adietish:
> Why do we have such file ? On vscode each extension is using the default one provided by vscode-commons

what file?

jeffmaury:
> 
> 
> > Why do we have such file ? On vscode each extension is using the default one provided by vscode-commons
> 
> what file?

segment.properties

adietish:
> > > Why do we have such file ? On vscode each extension is using the default one provided by vscode-commons
> > 
> > 
> > what file?
> 
> segment.properties

ok, removing it so that intellij-tekton is using the defaults in intellij-redhat-telemetry

adietish:
> > now sending the following events:
> > 
> > * "open editor" when an editor is opened:
> > 
> > ```
> >  "properties": {
> >     "app_name": "IntelliJ IDEA",
> >     "app_version": "2019.3",
> >     "duration": "00:00:00",
> >     "extension_name": "Tekton Pipelines by Red Hat",
> >     "extension_version": "0.7.0-SNAPSHOT",
> >     "java_version": "11.0.4",
> >     "resource_kind": "clustertasks",
> >     "resource_namespace": "default"
> >   },
> > ```
> 
@lstocchi 
> @jeffmaury about the open in editor action, would it make sense to have a field that say why the editor was opened? Like, editor opened to create new resource or opened to edit a task ?

I believe that this info is now gathered by the **resource_kind** and **resource_crud** properties:
```yaml
"properties": {
    "app_name": "IntelliJ IDEA",
    "app_version": "2019.3",
    "duration": "00:00:00",
    "extension_name": "Tekton Pipelines by Red Hat",
    "extension_version": "0.7.0-SNAPSHOT",
    "java_version": "11.0.4",
-->    "resource_crud": "create",
-->    "resource_kind": "ClusterTask",
    "resource_version": "tekton.dev/v1beta1",
    "result": "success"
  },
```
Fits your needs?

adietish:
I added a preferences page in Tekton that allows to switch to the Telementry preferences by clicking on the link:
image

jeffmaury:
Why is there a separate tekton telemetry page and not an addition to the existing one?

adietish:
> Why is there a separate tekton telemetry page and not an addition to the existing one?

looks like I didn't understand what you suggested. An addition to the existing one? Can you please explain?

jeffmaury:
Yes Tekton plugin already has a preference page so telemetry information that link to core telemetry settings should be in Tekton preferences. 
On the left, I see `Tekton Pipelines by Red Hat` and `Tekton Pipelines Telemetry`, there should be only `Tekton Pipelines by Red Hat`

adietish:
@jeffmaury added the text and link to the existing page:
image

lstocchi:
Played with it and looks good. A couple of thoughts.
The first is about `open in editor` action. We have a setting that allows users to redirect logs to the editor. 
image
It looks like a new `open in editor` action is triggered in telemetry every time a new line is added to the editor tab. Not a big issue but maybe we can prevent from adding it if the resource is empty? WDYT?
I ended up with this list for a small pipeline. Imagine having a long pipeline that run for minutes with tons of logs.
image

The difference with a normal `open in editor` action is that the resource kind is empty
image

lstocchi:
Second thought.
Couldn't it be useful to also track if someone changes settings? So we know if they are used or not?

lstocchi:
Then i'm a bit disappointed by woopra.
Most recent country: United States - timezone: Rome :cry: 
image

lstocchi:
@adietish can you please rebase? We refactored the tree and it only shows the current active namespace now. So i can test telemetry with it. Thanks!!

adietish:
> Played with it and looks good. A couple of thoughts.
> The first is about `open in editor` action. We have a setting that allows users to redirect logs to the editor.
> image
> It looks like a new `open in editor` action is triggered in telemetry every time a new line is added to the editor tab. Not a big issue but maybe we can prevent from adding it if the resource is empty? WDYT?
> I ended up with this list for a small pipeline. Imagine having a long pipeline that run for minutes with tons of logs.
> image
> 
> The difference with a normal `open in editor` action is that the resource kind is empty
> image

+1. I dont think that a telemetry log for every new line in the logs makes any sense.

adietish:
rebased, pushed force.

adietish:
> Then i'm a bit disappointed by woopra.
> Most recent country: United States - timezone: Rome 😢
> image

that's unexpected. I report the country by taking it from the timezone. Could you please verify what you get in (intellij-redhat-telemetry) `Environment.Builder.ensureCountry()`?

adietish:
@lstocchi removed duplicate telemetry notifications that occurred for editor opening actions (FollowLogs, ShowLogs) in 1cbef5a8fe2d8916ef7b6b96bc45fc57794534bc I moved the telemetry to the upper most location in the call hierarchy. This then required VirtualFileHelper to throw exceptions that then the caller would catch.
Please have a look if you agree.

lstocchi:
> @lstocchi are we sure that sonar works correctly? I see it always reporting 0% problems even though I see the local version of it complaining.

@jeffmaury any idea?

jeffmaury:
> 
> 
> > @lstocchi are we sure that sonar works correctly? I see it always reporting 0% problems even though I see the local version of it complaining.
> 
> @jeffmaury any idea?

I think Sonar is using the default mode which does not recognize as a full Java project. @sbouchet is working on IJ OpenShift and once this is good, we will report setting here

sbouchet:
> 
> I think Sonar is using the default mode which does not recognize as a full Java project. @sbouchet is working on IJ OpenShift and once this is good, we will report setting here

Sonar and gradle and GH Actions are kinda touchy to configure :D I'm close to have it works nicely, i'll ping/open a pr when I'm 100% sure it works on ij-openshift.

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/pipeline/CreatePipelineAction.java
```diff
@@ -13,26 +13,50 @@
...
         String content = getSnippet("Tekton: Pipeline");
 
-        if (!Strings.isNullOrEmpty(content)) {
-            VirtualFileHelper.createAndOpenVirtualFile(anActionEvent.getProject(), namespace,namespace + "-newpipeline.yaml", content, KIND_PIPELINES, item);
+        ActionMessage telemetry = TelemetryService.instance()
+                .action("create pipeline");
+        if (Strings.isNullOrEmpty(content)) {
+            telemetry.error("snippet content empty").send();
+        } else {
+            telemetry.send();
```
lstocchi, line 51: Should this be removed?

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/triggers/CreateTriggerTemplateAction.java
```diff
@@ -13,25 +13,49 @@
...
         String content = getSnippet("Tekton: TriggerTemplate");
+        ActionMessage telemetry = TelemetryService.instance()
+                .action("create trigger template");
 
-        if (!Strings.isNullOrEmpty(content)) {
-            VirtualFileHelper.createAndOpenVirtualFile(anActionEvent.getProject(), namespace, namespace + "-newtriggertemplate.yaml", content, KIND_TRIGGERTEMPLATES, item);
+        if (Strings.isNullOrEmpty(content)) {
+            telemetry.error("snippet content empty").send();
+        } else {
+            telemetry.send();
```
lstocchi, line 50: Should this be removed?

src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCli.java
```diff
@@ -75,8 +78,13 @@
 import static com.redhat.devtools.intellij.tektoncd.Constants.FLAG_WORKSPACE;
 import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_PIPELINERUN;
 import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_TASKRUN;
+import static com.redhat.devtools.intellij.tektoncd.telemetry.TelemetryService.PROP_RESOURCE_KIND;
+import static com.redhat.devtools.intellij.telemetry.core.service.TelemetryMessageBuilder.*;
```
lstocchi, line 82: wildcard

lstocchi (commented):
Just gave a quick look at the code and looks good to me. Added a couple of comments.
I'll play with it tomorrow morning. 

BTW thanks for all the enhancements you made to the old code. Really appreciated :pray:

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/pipeline/CreatePipelineAction.java
adietish: if missing/empty snippets if nothing we'd want to report, yes. I was inserting this because I thought that we'd report that the action failed. Thoughts?

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/triggers/CreateTriggerTemplateAction.java
adietish: see comments above.

adietish:
@lstocchi I consider this implementation as completed. Please review.

lstocchi (commented):
The code looks good. Only a comment below. Start playing with it

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/pipeline/CreatePipelineAction.java
```diff
@@ -13,26 +13,55 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(CreatePipelineAction.class);
+
+    public CreatePipelineAction() {
+        super(PipelinesNode.class);
+    }
 
     @Override
     public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, Tkn tkncli) {
+        ActionMessageBuilder telemetry = instance().action("create pipeline")
+                .property(PROP_RESOURCE_KIND, KIND_CONDITIONS);
```
lstocchi, line 45: Here there should be `KIND_PIPELINE` instead of `KIND_CONDITION`

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/pipeline/CreatePipelineAction.java
```diff
@@ -13,26 +13,55 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.swing.tree.TreePath;
 
+import java.io.IOException;
+
+import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_CONDITIONS;
 import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_PIPELINES;
+import static com.redhat.devtools.intellij.tektoncd.telemetry.TelemetryService.*;
```
lstocchi, line 29: wildcard

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/triggers/CreateTriggerBindingAction.java
```diff
@@ -13,25 +13,53 @@
...
-            VirtualFileHelper.createAndOpenVirtualFile(anActionEvent.getProject(), namespace, namespace + "-newtriggerbinding.yaml", content, KIND_TRIGGERBINDINGS, item);
+        if (Strings.isNullOrEmpty(content)) {
+            telemetry
+                    .error("snippet content empty.")
+                    .send();
+        } else {
+            String name = namespace + "-newtriggerbinding.yaml";
+            try {
+                VirtualFileHelper.createAndOpenVirtualFile(anActionEvent.getProject(), namespace, name, content, KIND_TRIGGERBINDINGS, item);
+                telemetry.send();
```
lstocchi, line 56: this should be `telemetry.success().send()`

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/triggers/CreateTriggerTemplateAction.java
```diff
@@ -13,25 +13,55 @@
...
-            VirtualFileHelper.createAndOpenVirtualFile(anActionEvent.getProject(), namespace, namespace + "-newtriggertemplate.yaml", content, KIND_TRIGGERTEMPLATES, item);
+        if (Strings.isNullOrEmpty(content)) {
+            telemetry
+                    .error("snippet content empty")
+                    .send();
+        } else {
+            String name = namespace + "-newtriggertemplate.yaml";
+            try {
+                VirtualFileHelper.createAndOpenVirtualFile(anActionEvent.getProject(), namespace, name, content, KIND_TRIGGERTEMPLATES, item);
+                telemetry.send();
```
lstocchi, line 58: `telemetry.success().send()`

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/triggers/CreateEventListenerAction.java
```diff
@@ -13,25 +13,54 @@
...
-            VirtualFileHelper.createAndOpenVirtualFile(anActionEvent.getProject(), namespace, namespace + "-neweventlistener.yaml", content, KIND_EVENTLISTENERS, item);
+        if (Strings.isNullOrEmpty(content)) {
+            telemetry
+                    .error("snippet content empty")
+                    .send();
+        } else {
+            String name = namespace + "-neweventlistener.yaml";
+            try {
+                VirtualFileHelper.createAndOpenVirtualFile(anActionEvent.getProject(), namespace, name, content, KIND_EVENTLISTENERS, item);
+                telemetry.send();
```
lstocchi, line 57: `telemetry.success().send()`

lstocchi (changes requested):
Couple of thoughts.

The first is the missing telemetry inside the `saveInEditorListener` (here https://github.com/redhat-developer/intellij-tekton/blob/7cb52967f038081b5eb34330521477bad06102e0/src/main/java/com/redhat/devtools/intellij/tektoncd/listener/SaveInEditorListener.java#L85). When calling the `saveOnCluster`method this could fail bc the resource is not written correctly (e.g it misses the name/kind/apiversion/spec/invalid crd). I guess the telemetry should be added inside the catch and report such errors as well.

The second is about a new feature that can be added later on. Writing it here so we can talk about it. When someone install a task from the tektonHub we don't have anything in the telemetry (we only have tektonhub dialog was opened). Wouldn't it be possible to also send the name of the tasks installed? This shouldn't be covered by privacy, no? I mean the tasks are from the hub...

lstocchi:
I checked why for woopra i live in US
My machine is set this way 
image

and the telemetry plugin gets the country from this. So it's ok i would say.
But i have a question regarding the IP because i'm missing a piece i think. If the telemetry plugin uses the IP to get the country and then only send the country name to woopra, where is the problem? How do we violate user's privacy?

adietish:
> I checked why for woopra i live in US
> My machine is set this way
> image

ok, explains it. Thanks for bringing up and resolving this issue.

> and the telemetry plugin gets the country from this. So it's ok i would say.
> But i have a question regarding the IP because i'm missing a piece i think. If the telemetry plugin uses the IP to get the country and then only send the country name to woopra, where is the problem? How do we violate user's privacy?

telemetry is not using the IP to determine the country: 
* we're not allowed to query external geolocation-services in IJ (to map an IP to a country), we're even sending bogus empty IP's (https://github.com/redhat-developer/intellij-redhat-telemetry/blob/main/src/main/java/com/redhat/devtools/intellij/telemetry/core/service/segment/SegmentBroker.java#L175):
```
    .pair(PROP_IP, VALUE_NULL_IP)
```
* we use the timezone to determine the country as a best possible guess (https://github.com/redhat-developer/intellij-redhat-telemetry/blob/main/src/main/java/com/redhat/devtools/intellij/telemetry/core/service/Environment.java#L140):
```
    private void ensureCountry() {
        if (this.country == null) {
            /*
            * We're not allowed to query 3rd party services to determine the country.
            * Segment won't report countries for incoming requests.
            * We thus currently dont have any better solution than use the country in the Locale.
            */
            country(Locale.getDefault().getDisplayCountry());
        }
    }
```

adietish:
filed a #19 where the country is deduced from the timezone to fix the shortcoming that I found in https://github.com/redhat-developer/intellij-tekton/pull/341#issuecomment-801021213.

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/triggers/CreateTriggerBindingAction.java
adietish: `success()` is not required any longer. The builder would add it automatically if there's no result or error when sending:
https://github.com/redhat-developer/intellij-redhat-telemetry/blob/main/src/main/java/com/redhat/devtools/intellij/telemetry/core/service/TelemetryMessageBuilder.java#L179
```
        private void ensureResultOrError() {
            if (!hasProperty(PROP_ERROR)
                    && !hasProperty(PROP_RESULT)) {
                success();
            }
        }
```

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/triggers/CreateEventListenerAction.java
adietish: same here, see comment above.

adietish:
> The first is the missing telemetry inside the `saveInEditorListener` (here
> 
> https://github.com/redhat-developer/intellij-tekton/blob/7cb52967f038081b5eb34330521477bad06102e0/src/main/java/com/redhat/devtools/intellij/tektoncd/listener/SaveInEditorListener.java#L85
> ). When calling the `saveOnCluster`method this could fail bc the resource is not written correctly (e.g it misses the name/kind/apiversion/spec/invalid crd). I guess the telemetry should be added inside the catch and report such errors as well.

There's already telemetry in `DeployHelper#saveOnCluster`:
```java
        ActionMessage telemetry = TelemetryService.instance().action("save to cluster")
                .property(PROP_RESOURCE_KIND, model.getKind())
                .property(PROP_RESOURCE_VERSION, model.getApiVersion());
        try {
            String resourceNamespace = CRDHelper.isClusterScopedResource(model.getKind()) ? "" : namespace;
            boolean isNewResource = executeTkn(namespace, resourceNamespace, yaml, updateLabels, model, tknCli);
            telemetry.property(PROP_RESOURCE_CRUD, (isNewResource ? "create" : "update")).success()
                    .send();
        } catch (KubernetesClientException e) {
            String errorMsg = createErrorMessage(model, e);
            telemetry
                    .error(anonymizeResource(model.getName(), namespace, errorMsg))
                    .send();
```
Is this not sufficient? 

I added it in `DeployHelper` instead of `HubModel#installHubItem` and `SaveInEditorListener`. In the meantime I'd prefer the latter because of the insight that telemetry should be at the most specific place (and not in generic places). In that sense I'd agree on moving telemetry out of `DeployHelper`?

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/triggers/CreateTriggerTemplateAction.java
adietish: #success not required any more. One might argue on being more explicit though.

lstocchi:
> Is this not sufficient?

My concern is if the yaml is not valid. The `isValid` method (first row of `DeployHelper#saveOnCluster`) check the yaml file and throws an error if that is not valid (e.g it doesn't have a valid kind/name). If this happens the telemetry doesn't report anything.

adietish:
> > Is this not sufficient?
> 
> My concern is if the yaml is not valid. The `isValid` method (first row of `DeployHelper#saveOnCluster`) check the yaml file and throws an error if that is not valid (e.g it doesn't have a valid kind/name). If this happens the telemetry doesn't report anything.

Fixed it: 
And then here comes the other reason why I couldn't add telemetry to `SaveInEditorListener`: the model, which determines the kind of resource, validity etc. only exists in the `DeployHelper`

lstocchi (approved):
Played with it and it looks like all actions and messages are good. LGTM!!

lstocchi:
Now i'm from italy :laughing: 
image

adietish:
@lstocchi before closing this, we need the checks to pass. Isn't our publishing job not pushing intellij-redhat-telemetry to the repos?

adietish:
> Now i'm from italy 😆
> image

awesome, thanks to your insistence I found out that the country was still wrong. Thanks!

adietish:
@jeffmaury can we merge this?

jeffmaury:
> 
> 
> @jeffmaury can we merge this?

Need to review it

adietish:
rebased & pushed force.

USAGE_DATA.md
```diff
@@ -0,0 +1,17 @@
...
+    * action duration time
+    * action error message (in case of exception)
+    * action specific data if applicable (see details below)
+* when plugin is shut down
+
+### Actions operating on tekton resources
+* resource kind
+* resource schema version
+* creation/update of a resource
+* namespace supports triggers
```
jeffmaury, line 17: Not quite clear to me, what does it mean exactly

build.gradle
```diff
@@ -43,6 +52,8 @@ dependencies {
     //to explicitely declare it here. If the Tekton F8 is
     //updated check the version
     compile 'com.squareup.okio:okio:1.15.0'
+    // telemetry contributes annotations 13.0.0, so we need to declare newer version
+    compile 'org.jetbrains:annotations:16.0.1'
```
jeffmaury, line 56: That seems very ugly and error prone, we should not rely on those dependencies

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/AddTriggerAction.java
```diff
@@ -81,13 +82,16 @@ public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object s
                 }
 
                 String kind = (element instanceof PipelineNode) ? "Pipeline " : "Task ";
+                telemetry.property(TelemetryService.PROP_RESOURCE_KIND, kind);
                 AddTriggerWizard addTriggerWizard = openTriggerBindingWizard(anActionEvent, element, triggerBindingTemplates, model, kind);
                 if (!addTriggerWizard.isOK()) {
+                    telemetry.result("wizard aborted")
```
jeffmaury, line 88: Shouldn't we have a constant to report abort because I think it's not specific to this action

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/CancelAction.java
```diff
@@ -52,22 +56,28 @@ public boolean isVisible(Object selected) {
 
     @Override
     public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, Tkn tkncli) {
+        ActionMessage telemetry = TelemetryService.instance().action("cancel");
         ExecHelper.submit(() -> {
             ParentableNode element = getElement(selected);
             String namespace = element.getNamespace();
             try {
                 if (element instanceof PipelineRunNode) {
+                    telemetry.property(TelemetryService.PROP_RESOURCE_KIND, Constants.KIND_PIPELINERUN).send();
```
jeffmaury, line 65: Should be after the cli call otherwise action will be sent twice

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/CancelAction.java
```diff
@@ -52,22 +56,28 @@ public boolean isVisible(Object selected) {
...
                     tkncli.cancelPipelineRun(namespace, element.getName());
                 } else if (element instanceof TaskRunNode) {
+                    telemetry.property(TelemetryService.PROP_RESOURCE_KIND, Constants.KIND_TASKRUN).send();
```
jeffmaury, line 68: Should be after the cli call otherwise action will be sent twice

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/CancelAction.java
```diff
@@ -52,22 +56,28 @@ public boolean isVisible(Object selected) {
...
+                telemetry
+                        .error(anonymizeResource(element.getName(), namespace, e.getMessage()))
+                        .send();
                 Notification notification = new Notification(NOTIFICATION_ID,
                         "Error",
                         element.getName() + " in namespace " + namespace + " failed to cancel\n" + e.getLocalizedMessage(),
                         NotificationType.ERROR);
                 Notifications.Bus.notify(notification);
-                logger.warn("Error: " + e.getLocalizedMessage(), e);
+                logger.warn("Error: " + e.getLocalizedMessage());
```
jeffmaury, line 80: Why did you remove the exception, stack trace will be missing from log, no ?

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/RefreshAction.java
```diff
@@ -30,6 +30,9 @@
 
 import javax.swing.tree.TreePath;
 
+import static com.redhat.devtools.intellij.tektoncd.telemetry.TelemetryService.*;
```
jeffmaury, line 33: Starred imports are prohibited

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/ShowDiagnosticDataAction.java
```diff
@@ -14,39 +14,60 @@
...
+                    telemetry.property(PROP_RESOURCE_KIND, Constants.KIND_PIPELINERUN);
                     hasDataToShow = tkncli.getDiagnosticData(namespace, "tekton.dev/pipelineRun", element.getName());
                 } else if (element instanceof TaskRunNode) {
+                    telemetry.property(PROP_RESOURCE_KIND, Constants.KIND_TASKRUN);
                     hasDataToShow = tkncli.getDiagnosticData(namespace, "tekton.dev/taskRun", element.getName());
                 }
                 if (!hasDataToShow) {
+                    String errorMessage = "No data available for " + element.getName() + " in namespace " + namespace + ".";
+                    telemetry
+                            .error(anonymizeResource(element.getName(), namespace, errorMessage))
```
jeffmaury, line 56: Not sure this has to be considered as an error (historical data missing from cluster) maybe @lstocchi can comment

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/StartAction.java
```diff
@@ -48,63 +49,79 @@
...
-                    String titleDialog = ((element instanceof PipelineNode)? "Pipeline " : "Task ") + element.getName();
+                StartWizard startWizard = UIHelper.executeInUI(() -> {
+                    String titleDialog = ((element instanceof PipelineNode) ? "Pipeline " : "Task ") + element.getName();
                     StartWizard wizard = new StartWizard(titleDialog, element, getEventProject(anActionEvent), model);
                     wizard.show();
                     return wizard;
                 });
+                if (startWizard != null && !startWizard.isOK()) {
+                    telemetry
+                            .result("wizard aborted")
```
jeffmaury, line 96: Constant welcomed

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/TektonAction.java
```diff
@@ -73,7 +73,7 @@ public String getSnippet(String snippet) {
         try {
             content = SnippetHelper.getBody(snippet);
         } catch (IOException e) {
-            logger.warn("Error: " + e.getLocalizedMessage(), e);
+            logger.warn("Error: " + e.getLocalizedMessage());
```
jeffmaury, line 76: Exception removal

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/logs/FollowLogsAction.java
```diff
@@ -10,35 +10,73 @@
...
         } catch (IOException e) {
+            telemetry
+                    .error(anonymizeResource(resourceName, namespace, e.getMessage()))
+                    .send();
             UIHelper.executeInUI(() ->
                     Messages.showErrorDialog(
                             "An error occurred while requesting logs for " + resourceName + "\n" + e.getLocalizedMessage(),
                             "Error"));
-            logger.warn("Error: " + e.getLocalizedMessage(), e);
+            logger.warn("Could not follow logs: " + e.getLocalizedMessage());
```
jeffmaury, line 74: Exception removal

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/logs/ShowLogsAction.java
```diff
@@ -12,40 +12,61 @@
...
+            telemetry
+                    .error(anonymizeResource(resourceName, namespace, errorMessage))
+                    .send();
             UIHelper.executeInUI(() ->
                     Messages.showErrorDialog(
-                            "An error occurred while requesting logs for " + resourceName + "\n" + e.getLocalizedMessage(),
+                            errorMessage,
                             "Error"));
-            logger.warn("Error: " + e.getLocalizedMessage(), e);
+            logger.warn(errorMessage);
```
jeffmaury, line 64: Exception removal

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/pipeline/CreatePipelineAction.java
```diff
@@ -13,26 +13,55 @@
...
+        } else {
+            String name = namespace + "-newpipeline.yaml";
+            try {
+                VirtualFileHelper.createAndOpenVirtualFile(anActionEvent.getProject(), namespace, name, content, KIND_PIPELINES, item);
+                telemetry.send();
+            } catch (IOException e) {
+                telemetry
+                        .error(anonymizeResource(name, namespace, e.getMessage()))
+                        .send();
+                logger.warn("Could not create resource: " + e.getLocalizedMessage());
```
jeffmaury, line 63: Exception removal

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/resource/CreateResourceAction.java
```diff
@@ -12,27 +12,52 @@
...
+        } else {
+            String name = namespace + "-newresource.yaml";
+            try {
+                VirtualFileHelper.createAndOpenVirtualFile(anActionEvent.getProject(), namespace, name, content, KIND_RESOURCES, item);
+                telemetry.success().send();
+            } catch (IOException e) {
+                telemetry
+                        .error(anonymizeResource(name, namespace, e.getMessage()))
+                        .send();
+                logger.warn("Could not create resource: " + e.getLocalizedMessage());
```
jeffmaury, line 59: Exception removal

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/task/CreateTaskAction.java
```diff
@@ -13,25 +13,52 @@
...
+        } else {
+            String name = namespace + "-newtask.yaml";
+            try {
+                VirtualFileHelper.createAndOpenVirtualFile(anActionEvent.getProject(), namespace, name, content, KIND_TASKS, item);
+                telemetry.send();
+            } catch (IOException e) {
+                telemetry
+                        .error(anonymizeResource(name, namespace, e.getMessage()))
+                        .send();
+                logger.warn(e.getLocalizedMessage());
```
jeffmaury, line 60: Exception removal

jeffmaury (commented):
Comments inline.
Also, I think it would be good to better qualify actions names:
- related to actions actions.xxxx
- other other.xxx or something else

USAGE_DATA.md
adietish: Reporting trigger support was removed. The entry in usage data should be removed.

adietish: resolving this conversation. If there's more to it, feel free to reopen it.

build.gradle
adietish: what alternative do you suggest? Without it, compilation fails with the following error:
```
> Task :compileJava
/Users/andredietisheim/Documents/jboss-workspaces/intellij-tekton/src/main/java/com/redhat/devtools/intellij/tektoncd/settings/SettingsConfigurable.java:24: error: cannot find symbol
    public @Nls(capitalization = Nls.Capitalization.Title) String getDisplayName() {
                ^
  symbol:   method capitalization()
  location: @interface Nls
/Users/andredietisheim/Documents/jboss-workspaces/intellij-tekton/src/main/java/com/redhat/devtools/intellij/tektoncd/settings/SettingsConfigurable.java:24: error: cannot find symbol
    public @Nls(capitalization = Nls.Capitalization.Title) String getDisplayName() {
                                    ^
  symbol:   variable Capitalization
  location: @interface Nls
Note: /Users/andredietisheim/Documents/jboss-workspaces/intellij-tekton/src/main/java/com/redhat/devtools/intellij/tektoncd/listener/TreeDoubleClickListener.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
2 errors

> Task :compileJava FAILED
```

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/AddTriggerAction.java
adietish: can add a constant but it is imho of limited use. The value is reported as value for the "result" property. Can go for "aborted" instead?

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/RefreshAction.java
adietish: weird, even though I have configured IJ to only introduce those once I have 999 imports, it keeps happening.

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/ShowDiagnosticDataAction.java
lstocchi: I missed it before but yeah, you're right. This is not an actual error. It may happen there are no data

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/logs/ShowLogsAction.java
```diff
@@ -12,40 +12,61 @@
...
 import com.redhat.devtools.intellij.tektoncd.tree.PipelineRunNode;
 import com.redhat.devtools.intellij.tektoncd.tree.TaskNode;
 import com.redhat.devtools.intellij.tektoncd.tree.TaskRunNode;
 import java.io.IOException;
+
+import com.redhat.devtools.intellij.telemetry.core.service.TelemetryMessageBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static com.redhat.devtools.intellij.tektoncd.Constants.*;
```
jeffmaury, line 30: Starred import

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/DeployHelper.java
```diff
@@ -28,26 +29,47 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static com.redhat.devtools.intellij.tektoncd.telemetry.TelemetryService.*;
```
jeffmaury, line 32: Starred import

jeffmaury (changes requested):

build.gradle
adietish: @jeffmaury thoughts?

adietish:
@jeffmaury see my comments above about the annotations that details why this import seems required

adietish:
> Comments inline.
> Also, I think it would be good to better qualify actions names:
> 
>     * related to actions actions.xxxx
> 
>     * other other.xxx or something else

ideally, this should be done in intellij-redhat-telemetry

jeffmaury:
OK seems an issue with dependency resolution but I think this dependency should be in compile scope so that it does not end up in the zip file

adietish:
> OK seems an issue with dependency resolution but I think this dependency should be in compile scope so that it does not end up in the zip file

isn't it in compile scope as you suggest?
```
compile 'org.jetbrains:annotations:16.0.1'
```

jeffmaury:
> 
> 
> > Comments inline.
> > Also, I think it would be good to better qualify actions names:
> > ```
> > * related to actions actions.xxxx
> > 
> > * other other.xxx or something else
> > ```
> 
> ideally, this should be done in intellij-redhat-telemetry

I don't see how redhat-telemetry could guess this domain specific prefixes unless we agree on them which according to telemetry stakeholders is not yet ready

jeffmaury:
> 
> 
> > I don't see how redhat-telemetry could guess this domain specific prefixes unless we agree on them which according to telemetry stakeholders is not yet ready
> 
> my idea was to add a prefix when you call a method:
> 
> ```
> telemetry.action("start last run)"
> ```
> 
> Would create an event name "action: start last run". Missing something?
> I can of course manually add those prefixes in intellij-tekton but would then have to remove them again once the change lands in intellij-redhat-telemetry?

How IJ telemetry will decide which prefix to use ?

adietish:
> > > I don't see how redhat-telemetry could guess this domain specific prefixes unless we agree on them which according to telemetry stakeholders is not yet ready
> > 
> > 
> > my idea was to add a prefix when you call a method:
> > ```
> > telemetry.action("start last run)"
> > ```
> > 
> > 
> > Would create an event name "action: start last run". Missing something?
> > I can of course manually add those prefixes in intellij-tekton but would then have to remove them again once the change lands in intellij-redhat-telemetry?
> 
> How IJ telemetry will decide which prefix to use ?

based to the method called:

* `telementry.action("some name")` would add the prefix "action:" to the name "some name" -> "action: some name"
* `telementry.other("some name")` would add the prefix "other:" to the name "some name" -> "other: some name"

missing something?

jeffmaury:
> 
> 
> > > > I don't see how redhat-telemetry could guess this domain specific prefixes unless we agree on them which according to telemetry stakeholders is not yet ready
> > > 
> > > 
> > > my idea was to add a prefix when you call a method:
> > > ```
> > > telemetry.action("start last run)"
> > > ```
> > > 
> > > 
> > > Would create an event name "action: start last run". Missing something?
> > > I can of course manually add those prefixes in intellij-tekton but would then have to remove them again once the change lands in intellij-redhat-telemetry?
> > 
> > 
> > How IJ telemetry will decide which prefix to use ?
> 
> based to the method called:
> 
>     * `telementry.action("papa smurf")` would add the prefix "action:" to the name -> "action: papa smurf"
> 
>     * `telementry.other("smurfette")` would add the prefix "other:" to the name -> "other: smurfette"
> 
> 
> missing something?

That is not scalable as it requires API change each time a new domain comes in and it makes the API more and more complex when times goes on

adietish:
@jeffmaury 
> That is not scalable as it requires API change each time a new domain comes in and it makes the API more and more complex when times goes on

added "action: "-prefix to telemetry messages. So far there are only telemetry messages that are triggered by actions. Some of the messages are created further down in the call hierarchy (`TknCli` l552, l568, `DeployHelper` l47, `TreeHelper` l172) but are still triggered by a single action. 
Alternatively one could also use the classname instead of some string.

adietish:
> LGTM.
> Remarks:
> 
>     * I would not have chosen the action: prefix when we push to the cluster because we may have several occurrences of it (each time user saves the file)
what do you have in mind as alternative? 
>     * still the diagnostics handler reports an error when there is no data found
can you explain? I dont get what you mean here?

jeffmaury:
> 
> 
> > ```
> > * I would not have chosen the action: prefix when we push to the cluster because we may have several occurrences of it (each time user saves the file)
> > ```
> 
> what do you have in mind as alternative?
io: ? storage: ?

> 
> > ```
> > * still the diagnostics handler reports an error when there is no data found
> > ```
> 
> can you explain? I dont get what you mean here?
See the discussion around the ShowDiagnosticsDataAction

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/ShowDiagnosticDataAction.java
```diff
@@ -14,44 +14,66 @@
...
+                    telemetry.property(PROP_RESOURCE_KIND, Constants.KIND_PIPELINERUN);
                     hasDataToShow = tkncli.getDiagnosticData(namespace, "tekton.dev/pipelineRun", element.getName());
                 } else if (element instanceof TaskRunNode) {
+                    telemetry.property(PROP_RESOURCE_KIND, Constants.KIND_TASKRUN);
                     hasDataToShow = tkncli.getDiagnosticData(namespace, "tekton.dev/taskRun", element.getName());
                 }
                 if (!hasDataToShow) {
+                    String message = "No data available for " + element.getName() + " in namespace " + namespace + ".";
+                    telemetry
+                            .result(anonymizeResource(element.getName(), namespace, message))
```
adietish, line 57: @jeffmaury, @lstocchi : corrected sending "no data" from error -> result

lstocchi: :+1: merge merge merge merge

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/ShowDiagnosticDataAction.java
adietish: corrected this, now sending it as #result

adietish:
Here are all the telemetry events that are sent:
image

I see 4 categories:
* events that report creating, editing, deleting resources -> "io:" or "crud:"
** CreateXXAction
** DeleteAction
** DeployHelper ("save to cluster")
* events that report starting resources -> "start/stop:"
** MirrorStartAction
** StartLastRunAction
**  StartAction
* events that report looking at the logs, diagnostic: -> "logs:"
** FollowLogsAction
** ShowLogsAction
** ShowDiagnosticDataAction
* miscellaneous events: -> "misc:" 
** CancelAction
** RefreshAction

@lstocchi, @jeffmaury: thoughts?

adietish:
looks as if the events that we send with a prefix are not listed in the schemas in woopra. Sent woopra support a request.

lstocchi:
> * events that report creating, editing, deleting resources -> **"io:"** or **"crud:"**
> * events that report starting resources -> **"start/stop:"**
> * events that report looking at the logs, diagnostic: -> **"logs:"**
> * miscellaneous events: -> **"misc:"**
> @lstocchi, @jeffmaury: thoughts?

Looks good to me.

lstocchi:
> looks as if the events that we send with a prefix are not listed in the schemas in woopra. Sent woopra support a request.

What schemas are you talking about? The activity report?

adietish:
@lstocchi yes, in woopra, the events with a prefix are listed in the activity, but not in the action schemas (configure > Action Schema)

adietish:
woopra support basically said that the events would not create schemas automatically because the ":" is an illegal character for them. Allowed are 

>lower case letters, numbers, underscore, dashes, and spaces

So we either change our naming scheme and use " ", "-", "_" or add the schemas manually.
I'd opt for "-"
@jeffmaury, @lstocchi: thoughts?

Here's the complete answer:
>When a new event without a schema comes into Woopra, the system will auto-generate a basic schema so that the event can be found in reports and searched for in the Woopra interface. However, for Schemas to be auto-generated, the allowed characters are -- lower case letters, numbers, underscore, dashes, and spaces. Since the actions contain a colon ":" they'll just have to be added to the schema manually since the character will prevent it from being automatically added.

jeffmaury:
> 
> 
> woopra support basically said that the events would not create schemas automatically because the ":" is an illegal character for them. Allowed are
> 
> > lower case letters, numbers, underscore, dashes, and spaces
> 
> So we either change our naming scheme and use " ", "-", "_" or add the schemas manually.
> I'd opt for "-"
Go for it
> 
> @jeffmaury, @lstocchi: thoughts?
> 
> Here's the complete answer:
> 
> > When a new event without a schema comes into Woopra, the system will auto-generate a basic schema so that the event can be found in reports and searched for in the Woopra interface. However, for Schemas to be auto-generated, the allowed characters are -- lower case letters, numbers, underscore, dashes, and spaces. Since the actions contain a colon ":" they'll just have to be added to the schema manually since the character will prevent it from being automatically added.

adietish:
@lstocchi, @jeffmaury: introduced the categories that we discussed above:
image
Please review, imho we're ready to merge, everything is done here.

lstocchi:
Looks good to me. it took me a while to understand the `Diag` label .... ah diagnostic :laughing:

adietish:
@jeffmaury good to go aka merge?

jeffmaury:
> 
> 
> @jeffmaury good to go aka merge?

I'm kinda lost here as I still see : as separator

adietish:
> > @jeffmaury good to go aka merge?
> 
> I'm kinda lost here as I still see : as separator

stupid me, fixing.

jeffmaury (approved):
LGTM

PR #352, opened 2021-02-18
feat(UI): refactor tree to only show current active namespace

### LINKED ISSUE #280: Display active namespace in tree view.

lstocchi:
it resolves #280 

@jeffmaury i had to change the modelsynchronizer, not sure if the new version should go to the common project as this is very customized to what we have in tekton.

Summary:

1. Only the current active namespace/project is shown
2. Watches are activated when nodes are loaded and they are always active (they are closed/removed only when the namespace/project is changed)


jeffmaury:
I tested it on my Minikube cluster and kubens tells me that I'm on the default namespace which has no Tekton resources at all. However, when I launched the IDE, I got: 

image

- the namespace node is not needed
- don't understand on which namespace you are

Need to review code now

lstocchi:
> * the namespace node is not needed
> * don't understand on which namespace you are

I don't know how you were able to have that tree lol. When i started my minikube it just showed my cluster node bc the namespace was null. BTW the name should be displayed in the namespace node, not sure why it's empty in your case. My doubt was if we wanted to have a first level with -> namespace, clustertask, clustertriggerbindings.. and then all other nodes below the namespace node. But i would keep the namespace node as it's easier to see which is the namespace the user is logged in. Where would you put the namespace name if you remove it?

src/main/java/com/redhat/devtools/intellij/tektoncd/tree/MutableTektonModelSynchronizer.java
```diff
@@ -0,0 +1,135 @@
+/*******************************************************************************
```
jeffmaury, line 1: Why is this needed ? Just wondering why we nee this Tekton specific version where almost 90% of the code is duplicated. At least, we should extends the common implementation

lstocchi:
it needs https://github.com/redhat-developer/intellij-common/pull/44 to work

jeffmaury:
According to our discussion during the last cabal, can you update this PR so that to remove the node corresponding to the current namespace and display the current namespace on the cluster node after the clusterURL (maybe separated by a slash)

lstocchi:
@jeffmaury now it's back to the namespace node. Give it a look whenever you can

jeffmaury (approved):
LGTM

PR #356, opened 2021-02-22
refactor: extracted method SaveInEditorListener#save

adietish:


lstocchi (approved):
LGTM

You just need to change the title of the PR to being complaint with conventional commits.
Here a list of all types https://github.com/pvdlg/conventional-commit-types
You can just rename this as `refactor: simplified (separation of concerns) SaveInEditorListener`

PR #362, opened 2021-02-23
refactor: extracted methods to reduce length of AddTriggerAction#actionPerformed

adietish:
extracted small methods and removed duplicate catching of IOException in order to reduce size of AddTriggerAction#actionPerformed (which is still quite large).


lstocchi (approved):
LGTM just change the PR title :wink:

PR #373, opened 2021-03-04
fix(UI): add title to tektonhub dialog (#357)

### LINKED ISSUE #357: Should have title in Tektonhub dialog (would help to know what this is all about)
**Steps:**
1. EXEC: in tool window tree: select "Tasks" category and open ctx menu
2. EXEC: pick "Import from Tekton Hub"
3. ASSERT: dialog with all available tasks in Tekton hub opens up
4. EXEC: look at the title bar, verify if there's a title

**Result:**
image
The dialog has no title. A title would help to know what this dialogs actually shows. Could be something like "Available Tekton Hub Tasks"

lstocchi:
it resolves #357


jeffmaury (approved):
LGTM

PR #376, opened 2021-03-04
fix(UI): change text in save dialog and showing kind and name of updated resource (#347)

### LINKED ISSUE #347: "Save to cluster" dialog doesn't tell me what resource is being saved
When saving a resource editor a dialog pops up and asks me if I want to save the resource to the cluster. 
image
It doesn't tell me anything about the resource that is saved. I can guess it by knowing about the current editor, the editor I was in when saving (or when it was triggered automatically - see #346).
I believe that we should at least mention the name and the type of the resource.

lstocchi:
it resolves #347


jeffmaury (approved):
LGTM

adietish (approved):

PR #378, opened 2021-03-04
fix(UI): optimize delete dialog (#332)

### LINKED ISSUE #332: Should improve dialog to confirm "Delete Resource"
When you delete resources you have to confirm that you really want to delete those by confirming the following dialog:
image

* The text for the checkbox **"Yes, delete the clustertask..."** is lengthy, repeats what's already said above and is not very clear: I didn't fully understand right away that it only concerns the related resources. 
Taking advice from the jetbrains style guide (which suggests using short and clear text at https://jetbrains.design/intellij/text/writing_short/) I'd maybe write just **Delete related resources**
* Same is true for **"The clustertask may also have..."**. 
I think that we can shorten it to **"Clustertask have related resources (ex. clustertaskRuns)"** or even drop it completely by adding the info to the checkbox label **"Delete resources related to clustertask [name]"**
* Same is true for **"Are you sure you want to delete...?"**. Jetbrains style guide says: "Remove verbs that explain the function of a UI control", "Do not address the user"
I would shorten it to **"Delete clustertask [name]?"**

lstocchi:
it resolves #332 

The message in the delete dialog is like this now
image

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/DeleteAction.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/DeleteAction.java
@@ -38,6 +38,11 @@ import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
 import javax.swing.tree.TreePath;
 
+
+import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_CLUSTERTASK;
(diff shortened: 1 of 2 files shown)


jeffmaury (approved):
LGTM

adietish (approved):

PR #394, opened 2021-03-29
feat(UI): add find task references action (#259)

### LINKED ISSUE #259: Action to find whether a component is actually used somewhere
Sometimes you are in a need to edit a task or delete it but you are unsure if it is used somewhere (e.g task inside a pipeline) in the cluster. It can be useful to have an action that searches for component references and tell us where we can find it.

lstocchi:
it resolves #259 .

This patch adds the action to find references of a task/clustertask inside the cluster. 
A tool window is opened in the bottom consisting of a tree that contains all resources that contains a reference to the task/ct. Each resource can be opened in read-only from there and also in the editor, in the classic way, using the `jump to source` action (as IJ does for its `Find Usages` action).

In the delete dialog there is a message to inform the user whenever the resource to be deleted is used somewhere.

findtaskref

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/DeleteAction.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/DeleteAction.java
@@ -180,7 +180,7 @@ public class DeleteAction extends TektonAction {
             List<RefUsage> usages = tkn.findTaskUsages(kind, name);
             return usages.size();
         } catch (IOException e) {
-            logger.warn(e.getLocalizedMessage());
+            logger.warn(e.getLocalizedMessage(), e);
             return 0;
         }
     }
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/DeleteAction.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/DeleteAction.java
@@ -33,26 +33,19 @@ import com.redhat.devtools.intellij.tektoncd.tree.TriggerTemplateNode;
 import com.redhat.devtools.intellij.tektoncd.ui.DeleteDialog;
 import com.redhat.devtools.intellij.tektoncd.ui.toolwindow.findusage.RefUsage;
 import com.redhat.devtools.intellij.tektoncd.utils.TreeHelper;
-
-import javax.swing.tree.TreePath;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
+import javax.swing.tree.TreePath;
 
+
+import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_CLUSTERTASK;
+import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_TASK;
+import static com.redhat.devtools.intellij.tektoncd.telemetry.TelemetryService.NAME_PREFIX_CRUD;
 import static com.redhat.devtools.intellij.telemetry.core.service.TelemetryMessageBuilder.ActionMessage;
 import static com.redhat.devtools.intellij.telemetry.core.util.AnonymizeUtils.anonymizeResource;
-import static com.redhat.devtools.intellij.tektoncd.telemetry.TelemetryService.NAME_PREFIX_CRUD;
-
-
-import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_CLUSTERTASK;
-import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_PIPELINE;
-import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_TASK;
-
-
-import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_CLUSTERTASK;
-import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_TASK;
 
 public class DeleteAction extends TektonAction {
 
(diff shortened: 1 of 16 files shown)


jeffmaury (changes requested):
Does not work for me as the cli `findTaskUsages` always reports an error. I changed 

`String jsonPathExpr = "jsonpath=\\\"{range .items[*]}{@.metadata.name}|{range .spec.tasks[*]}{.taskRef.kind},{.taskRef.name}|{end}{end}\\\"";
`
to make it work but not sure it is still working on MacOS/Linux

lstocchi:
> Does not work for me as the cli `findTaskUsages` always reports an error. I changed
> 
> `String jsonPathExpr = "jsonpath=\\\"{range .items[*]}{@.metadata.name}|{range .spec.tasks[*]}{.taskRef.kind},{.taskRef.name}|{end}{end}\\\""; `
> to make it work but not sure it is still working on MacOS/Linux

yes, it works fine on linux as well with your change.

jeffmaury (approved):
LGTM

lstocchi:
@jeffmaury added the telemetry thing. Can you give it a look before merging it? https://github.com/redhat-developer/intellij-tekton/pull/394/commits/b58f0189a10c688756be230e8dd5a23674c83965

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/DeleteAction.java
```diff
@@ -137,7 +118,71 @@ public void actionPerformed(AnActionEvent anActionEvent, TreePath[] path, Object
...
+    private int getUsages(Tkn tkn, String kind, String name) {
+        if (!(kind.equalsIgnoreCase(KIND_TASK) || kind.equalsIgnoreCase(KIND_CLUSTERTASK))) {
+            return 0;
+        }
+
+        try {
+            List<RefUsage> usages = tkn.findTaskUsages(kind, name);
+            return usages.size();
+        } catch (IOException e) {
+            logger.warn(e.getLocalizedMessage());
```
jeffmaury, line 183: Exception is missing

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/clustertask/FindClusterTaskRefAction.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected, Tkn tkncli) {
+        ParentableNode element = getElement(selected);
+        ExecHelper.submit(() -> {
+            try {
+                List<RefUsage> usages = tkncli.findTaskUsages(KIND_CLUSTERTASK, element.getName());
+                UIHelper.executeInUI(() -> FindTaskRefPanelBuilder.instance().build(anActionEvent.getProject(), KIND_CLUSTERTASK, element.getName(), usages));
+            } catch (IOException e) {
+                logger.warn(e.getLocalizedMessage());
```
jeffmaury, line 46: Same here

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/task/FindTaskRefAction.java
```diff
@@ -0,0 +1,59 @@
...
+        ExecHelper.submit(() -> {
+            try {
+                List<RefUsage> usages = tkncli.findTaskUsages(KIND_TASK, element.getName());
+                UIHelper.executeInUI(() -> FindTaskRefPanelBuilder.instance().build(anActionEvent.getProject(), KIND_TASK, element.getName(), usages));
+                telemetry.send();
+            } catch (IOException e) {
+                telemetry
+                        .error(anonymizeResource(null, element.getNamespace(), e.getMessage()))
+                        .send();
+                logger.warn(e.getLocalizedMessage());
```
jeffmaury, line 55: Same

src/main/java/com/redhat/devtools/intellij/tektoncd/ui/toolwindow/findusage/FindTaskRefPanelBuilder.java
```diff
@@ -0,0 +1,255 @@
...
+    private ActionListener getJumpSourceAction(Project project) {
+        return e -> {
+            LabelAndIconDescriptor node = getLastSelectedNode();
+            if (node != null && node.getElement() instanceof RefUsage) {
+                RefUsage ref = (RefUsage) node.getElement();
+                try {
+                    VirtualFile vf = TektonVirtualFileManager.getInstance(project).findResource(ref.getNamespace(), ref.getKind(), ref.getName());
+                    VirtualFileHelper.openVirtualFileInEditor(project, ref.getNamespace(), ref.getName(), ((LightVirtualFile) vf).getContent().toString(), ref.getKind(), false);
+                } catch (IOException ex) {
+                    logger.warn(ex.getLocalizedMessage());
```
jeffmaury, line 188: Same

jeffmaury (changes requested):

src/main/java/com/redhat/devtools/intellij/tektoncd/actions/DeleteAction.java
lstocchi: you mean `logger.warn(e.getLocalizedMessage(), e)`?

jeffmaury: Yes

jeffmaury (changes requested):
There a still a few calls to logger.warn without exception

lstocchi:
> There a still a few calls to logger.warn without exception

Sorry i missed those ones, done!

jeffmaury (approved):
LGTM

PR #395, opened 2021-03-29
chore: add dependency to kubernetes plugin (#370)

### LINKED ISSUE #370: Make IJ Tekton dependent on IJ k8s
As now we only show the current active namespace and the user is not able to change the namespace/cluster from the IDE itself, we need to make this plugin dependent on IJ kubernetes so users can use that one to switch

lstocchi:
it resolves #370 
I faced https://github.com/redhat-developer/intellij-kubernetes/issues/180, not sure if we want to wait until a fix for that is released

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/tree/TektonTreeStructure.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/tree/TektonTreeStructure.java
@@ -24,6 +24,7 @@ import com.redhat.devtools.intellij.common.utils.ExecHelper;
 import com.redhat.devtools.intellij.tektoncd.tkn.Run;
 import com.redhat.devtools.intellij.tektoncd.tkn.TaskRun;
 import com.redhat.devtools.intellij.tektoncd.tkn.Tkn;
+import com.redhat.devtools.intellij.tektoncd.tkn.TknCliFactory;
 import com.redhat.devtools.intellij.tektoncd.utils.WatchHandler;
 import io.fabric8.kubernetes.api.model.Config;
 import io.fabric8.kubernetes.api.model.NamedContext;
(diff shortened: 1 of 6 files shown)


jeffmaury (approved):
LGTM

lstocchi:
@jeffmaury updated by using the new IJ k8s which works fine. Now we can use it to switch cluster.

src/main/java/com/redhat/devtools/intellij/tektoncd/tree/TektonTreeStructure.java
```diff
@@ -526,7 +527,8 @@ private boolean hasNewToken(NamedContext newContext, Config newConfig, NamedCont
     protected void refresh() {
         try {
             WatchHandler.get().removeAll();
-            root.load();
+            initialized.set(false);
```
jeffmaury, line 530: Don't understand the rationale for that as refresh semantic is to refresh the tree but does not require the cluster connection to be reset ?

lstocchi: I have done this to prevent the tree from showing a wrong cluster url as root. 
When the user changes cluster, the tree is refreshed and the root is drawn again, but when this happens, the tkncli still has not been updated (the `initializeTkn` is slower) and the `tkn.getMasterUrl()` to retrieve the root name returns the old cluster url as it still points to the old cluster. So you end up having a tree that shows mixed results (values from old cluster until the tkncli update and then the correct ones).

lstocchi: I saw what you said regarding `LabelAndIconDescriptor` that accept a supplier but that doesn't solve the issue i have or maybe i don't see the way to use that. When the tree is refreshed, the `tkn cli` inside my supplier function will still point to the old cluster bc its update still is not complete.

An alternative (instead of using `initialized.set(false) ....`) could be to make `root.load()` returns a completablefuture and run `mutableModelSupport.fireModified(root)` when that is complete.

jeffmaury: I would prefer that approach as it's more or less what we have done in IJ OpenShift and it explains better the workflow

jeffmaury (approved):
LGTM

PR #403, opened 2021-04-13
chore: prepare 0.8.0

lstocchi:
I also added https://github.com/redhat-developer/intellij-tekton/pull/394 feature inside the readme. So after that one is approved, this can be reviewed as well.

Not sure if i need to add something else about the telemetry, let me know.


jeffmaury (changes requested):
I noticed build.gradle is referencing a SNAPSHOT version of intellij-common which we should not do for releases, so I will prepare a new release (that would be a 1.0.0 as there is a breaking change)

jeffmaury (approved):
LGTM

PR #408, opened 2021-05-05
feat(UI): add support for creating new PersistentVolumeClaim and/or VolumeClaimTemplate from start wizard and use them

### LINKED ISSUE #398: Support for volume claim templates for workspaces
When starting a pipeline or adding a trigger, user should be able to use a volume claim template for the workspaces as an alternative to creating a PVC:
https://tekton.dev/docs/pipelines/workspaces/#using-persistentvolumeclaims-as-volumesource

lstocchi:
it resolves #398 + also adds support for creating new PVC (something vscode already had and IJ misses)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,8 +38,8 @@ runIdeForUiTests {
 dependencies {
     compile 'io.fabric8:openshift-client:5.0.2'
     compile 'io.fabric8:tekton-client:5.0.2'
-    compile 'com.redhat.devtools.intellij:intellij-common:1.1.0'
-    testCompile 'org.mockito:mockito-inline:3.4.6'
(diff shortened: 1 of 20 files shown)


lstocchi:
I think I need to write some more tests but this should be ready to being used. 
Now it supports the creation of new vct and pvc directly from the start wizard and also the `mirror start` action is able to retrieve and show those data.

I followed what vscode does, so the UX is almost the same.

vct

jeffmaury (approved):
LGTM.
2 remarks:
- as the PR title will be reported in the change notes, I think PVC/VCT is not quite clear for users so worth replacing by a more explicit phrasing
- when creating a volume, units are displayed as expected by K8S API but I think it would be more user friendly to replace it with GB, TB, MB ?

jeffmaury (approved):
LGTM

PR #417, opened 2021-05-21
chore: Update schemas to Tekton Triggers 0.14.0

### LINKED ISSUE #413: Update schemas from Tekton Triggers 0.14.0

jeffmaury:
Fixes #413


lstocchi:
@jeffmaury LGTM but we still have a problem as the schema will report an error for any fields under resourcetemplates. This is the problem @sudhirverma has https://github.com/redhat-developer/vscode-tekton/issues/553

```
    "TriggerResourceTemplate": {
      "properties": {},
      "additionalProperties": false,
      "type": "object"
    },
```

lstocchi:
I don't know if we want to change the behavior of all fields that have the `properties` field equals to an empty project or not.  By a quick look it seems that the fields with empty properties are `Quantity`, `FieldsV1` and then some *Status... . If we set `additionalProperties` to true when `properties` is empty we should be fine. @jeffmaury  ??

jeffmaury:
Setting additionalProperties to true will allow any property thus defeat the validation. Need to understand why the properties list is empty

Issue #358, opened 2021-02-22
Not intuitive that "combo" triggers action when clicked in the white area


adietish:
**Steps:**
1. EXEC: in tool window tree: select "Tasks" category and open ctx menu
2. EXEC: pick "Import from Tekton Hub"
3. ASSERT: dialog with all available tasks in Tekton hub opens up and lists all available tasks (that one may install to the cluster)
4. ASSERT: next to an item there's a button labeled "install" that looks like a combo.
![image](https://user-images.githubusercontent.com/25126/108755792-f9435780-7547-11eb-9fbf-439d88c4756d.png)
5. EXEC: click the button (in the white area, not on the 2 arrows on the right)

**Result:**
The "combobox"-button would install, one can notice as clicking in results in a dialog appearing that asks if the task should be saved to the cluster.

**Expected result:**
The "combobox"-button should not look like a combo that one can click and discover more options. Clicking it results in installing the task to the cluster, not opening up a list of options.

jeffmaury:
That's on purpose because you can also install the task as cluster task

adietish:
> That's on purpose because you can also install the task as cluster task

Can you explain please? My point is that intuitively a user would expect a combo to show a list of options. Here, the button isn't a combo, it's a button that triggers an action. It thus think that it should look like a button.

adietish:
Ah, I grocked it, when clicking on the 2 arrows, it really is a combo:
![image](https://user-images.githubusercontent.com/25126/108766940-0bc48d80-7556-11eb-9ac3-4bfd92ecac72.png)
But when clicked on the white are it'll install directly. This wasn't obvious to me. I'll thus reformulate in the sense that it isn't obvious when the button acts like a combo and when as a button. 
In MacOS combos (that I ack look similar to this combo) list items when clicked anywhere, not just when clicked on the arrows on the right. It is thus confusing to MacOS users which dont expect this behaviour. Btw. the version combo in the details pane on the right behaves in this way. It'll list the options whereever you click.
I'd thus either change the behaviour (list the options regardless where one clicks - and then add a separate, explicit "Install" button in the details pane) or at least explain this somewhere in the dialog.

adietish:
Furthermore it isn't clear as what it would install when the 2nd option "Install" is selected:
![image](https://user-images.githubusercontent.com/25126/108768663-39aad180-7558-11eb-88bb-a896a72d59a9.png)
What does it do in this 2nd option?

[...]

PR #423, opened 2021-05-31
fix: remove border that does not overlap to combobox (#361)

### LINKED ISSUE #361: Install buttons in Tekton Hub dialog are not good looking when selected
**Steps:**
1. EXEC: select "ClusterTask", pick "Import from TektonHub" in ctx menu
2. ASSERT: list of available tasks in Tekton hub are listed
3. EXEC: select an item and look at the "install"-button

**Result:**
image
The button has rounded corners but an angular border around it. Between the 2 the background shines through. In my eyes this is not good looking.

lstocchi:
it fixes #361 and https://github.com/redhat-developer/intellij-tekton/issues/358

@adietish I changed how the buttons are created and now it is the same as in vscode. I don't use the combo anymore. Does it look better on your Mac? Is it easier to use?

hub

This is how it looks in vscode
image

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/InstallFromHubAction.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/InstallFromHubAction.java
@@ -32,7 +32,10 @@ public class InstallFromHubAction implements Action {
 
     @Override
     public Object getValue(String key) {
-        return this.text;
+        if (key.equals("Name")) {
+            return this.text;
+        }
+        return null;
     }
 
     @Override
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/InstallFromHubAction.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/actions/InstallFromHubAction.java
@@ -1,0 +1,67 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ * Red Hat, Inc.
+ ******************************************************************************/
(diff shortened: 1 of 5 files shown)


adietish (approved):
Much of an improvement to me, thus +1
Commented in https://github.com/redhat-developer/intellij-tekton/issues/361#issuecomment-857519424 and https://github.com/redhat-developer/intellij-tekton/issues/358#issuecomment-857535084 on why I consider this to be much better than what we had before. GJ!

ps. jetbrains design guide suggests to use a split button in those situations (or alternatively drop the option alltogether like @jeffmaury pointed out): https://jetbrains.design/intellij/controls/split_button/
Your buttons comes close to it and improves things.

lstocchi:
> ps. jetbrains design guide suggests to use a split button in those situations (or alternatively drop the option alltogether like @jeffmaury pointed out): https://jetbrains.design/intellij/controls/split_button/
> Your buttons comes close to it and improves things.

:scream: I didn't know of that split button. I'm going to use it then. The experience will be the same but atleast i'm going to clean the code a bit. Thanks for the tip!!

adietish:
nice, now fully IJ design guide compliant: 
image
+1

adietish (approved):
nice, now fully IJ design guide compliant: 
image
+1

ps. in my case hitting the button "Install as ClusterTask" takes ~5 seconds until the "Save to cluster" dialog shows up. I initially thought that my click wasn't noticed and thus started clicking on other buttons. But this delay might be my machine that's under heavy load of a minikube with plenty of pods running (knative & tekton extensions installed). A later enhancement could be to provide some visual feedback that the button was pressed. ex. turn the mouse pointer to a hour glass, disable the button or a progress bar etc. ?

lstocchi:
> ps. in my case hitting the button "Install as ClusterTask" takes ~5 seconds until the "Save to cluster" dialog shows up. I initially thought that my click wasn't noticed and thus started clicking on other buttons. But this delay might be my machine that's under heavy load of a minikube with plenty of pods running (knative & tekton extensions installed). A later enhancement could be to provide some visual feedback that the button was pressed. ex. turn the mouse pointer to a hour glass, disable the button or a progress bar etc. ?

Good point :+1:

PR #425, opened 2021-06-10
feat: add options to remove/restore yaml clutters (#309)

### LINKED ISSUE #309: Remove clutter from Yaml configurations
I came across this plugin that removes all useless things from a yaml output so to make the resource more readable https://github.com/itaysk/kubectl-neat and i was thinking if we should do something similar here. 

Like adding an option so the user can tell the plugin "when downloading and showing resources in the IDE editor, remove all stuff considered redundant as i don't care about them and make everything more readable". WDYT? @jeffmaury 

cc @adietish something that could work for the k8s plugin as well

lstocchi:
it resolves #309 

This patch allows to remove/restore yaml clutter.
It adds a setting so that user can decide if resource should be opened without clutters, and two commands within the editor's context menu

yaml-clutter


lstocchi:
@jeffmaury I need to write some tests but you should be able to test it and see if it misses something

jeffmaury (approved):
LGTM. However, I don't feel this is Tekton specific. So I see 2 options
- merge it and create an issue in IJ Common so as to share code
- delay merge and refactor so that code is shared for other plugins

lstocchi:
> LGTM. However, I don't feel this is Tekton specific. So I see 2 options
> 
> * merge it and create an issue in IJ Common so as to share code
> * delay merge and refactor so that code is shared for other plugins

Done, now it needs https://github.com/redhat-developer/intellij-common/pull/82 to work

jeffmaury (approved):
LGTM

PR #451, opened 2021-08-19
fix: update triggers to use beta1 (#450)

### LINKED ISSUE #450: Error while creating new eventListener, old schemas used

lstocchi:
it resolves #450


jeffmaury (commented):
LGTM. The only concern is that if force switches to v1beta1 for triggers which would not work for existings customers of OpenShift Pipelines.
So we have options:
- check if v1beta1 is available on the cluster
- delay this patch until v1beta1 is more deployed

lstocchi:
> LGTM. The only concern is that if force switches to v1beta1 for triggers which would not work for existings customers of OpenShift Pipelines.
> So we have options:
> 
> * check if v1beta1 is available on the cluster
> * delay this patch until v1beta1 is more deployed

@jeffmaury the problem is that it doesn't work with alpha1 as well. It's like we miss some update. If you try to create a trigger alpha1 with `name` property it fails. It only works with `ref`.
Now it searches for the beta1, if it's not found on the cluster it uses alpha1 as default.

Issue #452, opened 2021-08-23
Replace API currently used that are scheduled for removal in 2021.3


adietish:
relates to: https://github.com/redhat-developer/intellij-common/pull/94, https://github.com/redhat-developer/intellij-common/pull/91

lstocchi:
Closed by https://github.com/redhat-developer/intellij-tekton/pull/474

PR #453, opened 2021-08-23
use terminal widget to show logs (#452)

adietish:


adietish:
@jeffmaury can you please test this on Windows and verify that the log output does not produce special chars (that you reported on call)? 

You'd need to merge https://github.com/redhat-developer/intellij-common/pull/91 and publish common first. Then test this PR via "Show Log" action (the only one that's currently using the terminal widget replacement currently.
If it works we'd have a replacement for the broken API.

jeffmaury (changes requested):
I don't think this design work with several plugins (OpenShift,...)

adietish:
> I don't think this design work with several plugins (OpenShift,...)

What are the problems that you see?

PR #454, opened 2021-08-25
feat: show configurations in editor(#444)

### LINKED ISSUE #444: Add support for configMap used for tekton default values
Tekton uses the configMap `config-defaults` to set the default values to be used when running pipelines/tasks (e.g default service account, default pvc...). We should verify who can update this configMap (only admins?) and then allow users to edit it from the IDE.

lstocchi:
it resolves #444 

As agreed the configmaps are shown in read-only mode. 
I added two configmaps: config-defaults which contains the default values and feature-flags which contains the settings to enable alpha features like debug and bundles...

show-config


lstocchi:
@jeffmaury below how it is now
confignode

jeffmaury:
The content of the config maps is missing under each node.
See how the configuration is displayed in the Kafka plugin:

image

lstocchi:
Are we forced to do it because we want the same behavior across our plugins? Otherwise i don't see any reason of having such a UI. 

It could work because we mostly deal with small config files but if you have a lot of entries you force the user to scroll down the tree because a node has potentially many children, + if the value of the key-value pair is very long is not readable at all in the treeview and the user is forced to scroll horizontally or expand the panel completely, + our tree works that we pre-load children when the parent of their parent is expanded, so we have to download all configmaps even if they will not ever be read by the user. Opening the configMap into the editor is way better of listing everything in the tree imo, users can also use the search function for what they need and the commented part are visible as well.

E.g inside the config-defaults you have this data block, it is unreadable in the tree, you will have an _example node with a very long value.
```
data:
  _example: |
    ################################
    #                              #
    #    EXAMPLE CONFIGURATION     #
    #                              #
    ################################

    # This block is not actually functional configuration,
    # but serves to illustrate the available configuration
    # options and document them in a way that is accessible
    # to users that `kubectl edit` this config map.
    #
    # These sample configuration options may be copied out of
    # this example block and unindented to be in the data block
    # to actually change the configuration.

    # default-timeout-minutes contains the default number of
    # minutes to use for TaskRun and PipelineRun, if none is specified.
    default-timeout-minutes: "60"  # 60 minutes

    # default-service-account contains the default service account name
    # to use for TaskRun and PipelineRun, if none is specified.
    default-service-account: "default"

    # default-managed-by-label-value contains the default value given to the
    # "app.kubernetes.io/managed-by" label applied to all Pods created for
    # TaskRuns. If a user's requested TaskRun specifies another value for this
    # label, the user's request supercedes.
    default-managed-by-label-value: "tekton-pipelines"

    # default-pod-template contains the default pod template to use
    # TaskRun and PipelineRun, if none is specified. If a pod template
    # is specified, the default pod template is ignored.
    # default-pod-template:

    # default-cloud-events-sink contains the default CloudEvents sink to be
    # used for TaskRun and PipelineRun, when no sink is specified.
    # Note that right now it is still not possible to set a PipelineRun or
    # TaskRun specific sink, so the default is the only option available.
    # If no sink is specified, no CloudEvent is generated
    # default-cloud-events-sink:

    # default-task-run-workspace-binding contains the default workspace
    # configuration provided for any Workspaces that a Task declares
    # but that a TaskRun does not explicitly provide.
    # default-task-run-workspace-binding: |
    #   emptyDir: {}
```

jeffmaury:
The rule of thub is that is allows quick and easy access to the configuration. The usecase you mentionned can be handled with truncation

jeffmaury (approved):
Approving based on @adietish remark

PR #467, opened 2021-09-16
feat: Debug taskruns

### LINKED ISSUE #457: Add support for debug in Tekton
Example at https://jenkins-x.io/blog/2021/08/18/debug-tekton/

lstocchi:
it resolves #457 

This PR needs https://github.com/redhat-developer/intellij-common/pull/108 to work.

The changes made by this PR include:

- Changing some deprecated calls from fabric8 API because we switched to 5.7.2 using the new common
- Needed to switch from tkn cli to fabric8 when retrieving taskruns/pipelineruns (this includes changing to the tree structure) because the cli didn't return infos regarding *run started in debug mode 
- Everything needed to start a taskrun in debug mode

The only known issue is the terminal lagging, i'll keep trying to fix it but if everything works fine i think we can merge this PR and work on a new PR to fix the lagging problem


lstocchi:
Found out a problem with the tree, children of a pipelinerun failed to open in editor. Looking into it

jeffmaury:
> 
> 
> @jeffmaury edited. We had the same issue with watchers and polling. Give it a look please. Now I store the tkn instance per project within the factory so i can dispose it when the project requires a new tkn. WDYT?

TBH I don't think the factory has to keep a map I think the responsibility of release tkn instance resources should be on the owner (ie the tree/window) rather that the factory. That will make the factory much simpler with less side effects.

lstocchi:
> TBH I don't think the factory has to keep a map I think the responsibility of release tkn instance resources should be on the owner (ie the tree/window) rather that the factory. That will make the factory much simpler with less side effects.

My only concern (which is something i didn't cover yet ) is when you change project or close everything. Let's say you are using Project1 and you switch in same window to project2, the tree/window will be reloaded and they have no reference to the old tkn so you can't dispose it and some processes can remain there unused, no? If you give to the factory the responsability to keep track of what it creates you can check if a project is disposed and clean its resources.

jeffmaury (approved):
LGTM. Good stuff, sorry for the delay

PR #495, opened 2021-11-08
feat: start debug session from task (#470)

### LINKED ISSUE #470: Start debug session from task
Debug sessions can be started from an existing task run. We should be able to start a debug session from the task itself.
Reasons for that are:

- task runs may be too low level for novice users
- task runs may not already exist (new tasks)

lstocchi:
it resolves #470


src/main/java/com/redhat/devtools/intellij/tektoncd/actions/debug/DebugTaskAction.java
```diff
@@ -0,0 +1,69 @@
+package com.redhat.devtools.intellij.tektoncd.actions.debug;
```
jeffmaury, line 1: Missing header

jeffmaury (changes requested):

jeffmaury (approved):
LGTM

PR #496, opened 2021-11-15
fix: fix issue when running pipeline

lstocchi:
This patch solves some issues i found while running a complete pipeline. 
Bugs found:

- if a pipeline uses a clusterTask instead of a task, when trying to mirror that pipeline and restart it with same args, it failed. https://github.com/redhat-developer/intellij-tekton/pull/496/files#diff-0d5eb06ce6ad9596e824d43ccb769d677ca230960961ae03f6067c08006ab5aeR64

- there was an ambiguos recognition of an already created PVC and a PVC to be created before running a pipeline https://github.com/redhat-developer/intellij-tekton/pull/496/files#diff-860ef81cd8bd0707f909424161924aaed6c1a6adf357871caabc4cf07826f8c8R217

- there is an issue in fabric8 which make httpClient disposing when this is used by two clients (kubernetesclient and an adapted one). In our case we start with a kubernetesclient. In commons library an adapted openshiftclient is created for telemetry purpose (to retrieve ocp info) and then it's disposed. By disposing it, the httpClient used by the original kubernetClient is not usable anymore and tekton fails to open watches .... this is a workaround so we can release it. https://github.com/redhat-developer/intellij-tekton/pull/496/files#diff-0eb371265a17866a11e37696a7b5a1c0bf8cfd2d54229f74857c38f18dc00f38R159

- added a guard because i saw workspaces reset in some cases https://github.com/redhat-developer/intellij-tekton/pull/496/files#diff-6d03e68f4362750ec68c21b89813f85b9d38ccefbdb4d537e5941b9a74741e77R237


jeffmaury (approved):
LGTM

PR #499, opened 2021-11-21
fix: fix pvc/vct creation bug when starting pipeline or adding trigger

### LINKED ISSUE #501: PVC/VCT are not created correctly when starting pipeline or adding trigger
There are two kinds of bugs for the same thing.

Pvc is not created when starting a pipeline. It always create a vct.

Add trigger doesn't support creation for pvc/vct from wizard even if the option is available. When you try to create one, the code is missing. The ui is inherited from the start pipeline wizard

lstocchi:
it resolves #501

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/Constants.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/Constants.java
@@ -45,6 +45,7 @@ public class Constants {
     public static final String KIND_EVENTLISTENER = "eventlistener";
 
     public static final String KIND_PVC = "persistentVolumeClaim";
+    public static final String KIND_VCT = "volumeClaimTemplate";
     public static final String KIND_CONFIGMAP = "configMap";
     public static final String KIND_SECRET = "secret";
     public static final String KIND_EMPTYDIR =  "emptyDir";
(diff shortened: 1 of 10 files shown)


jeffmaury (commented):
My remarks:
- we should have a corresponding issue that describe the various use cases and what we are fixing with this
- when choosing Create new PVC or Create new VolumeClaimTemplate, the default proposed size is 0MB I think it should be 1MB
- when choosing Create new VolumeClaimTemplate, then the pipeline run definition has:

```
  spec:                                                 
    pipelineRef:                                        
      name: echo                                        
    serviceAccountName: default                         
    timeout: 1h0m0s                                     
    workspaces:                                         
    - name: repo                                        
      persistentVolumeClaim:                            
        claimName: Create new VolumeClaimTemplate       
```
That look strange to me

lstocchi:
> * when choosing Create new VolumeClaimTemplate, then the pipeline run definition has:
> 
> ```
>   spec:                                                 
>     pipelineRef:                                        
>       name: echo                                        
>     serviceAccountName: default                         
>     timeout: 1h0m0s                                     
>     workspaces:                                         
>     - name: repo                                        
>       persistentVolumeClaim:                            
>         claimName: Create new VolumeClaimTemplate       
> ```
> 
> That look strange to me

What did you do exactly? I'm not able to replicate it

jeffmaury:
tekton

lstocchi:
Can you please verify you are using this PR? I have the correct behavior by doing your exact steps

jeffmaury:
> Can you please verify you are using this PR? I have the correct behavior by doing your exact steps

This seems to be related to the default selection. So to reproduce you must:
- have no existing PVC
- only change the size of pvc template (if you don't that is the same)
- click start

lstocchi:
> > Can you please verify you are using this PR? I have the correct behavior by doing your exact steps
> 
> This seems to be related to the default selection. So to reproduce you must:
> 
> * have no existing PVC
> * only change the size of pvc template (if you don't that is the same)
> * click start

Thanks, replicated! :+1:

jeffmaury (approved):
LGTM

PR #507, opened 2021-11-29
feat: warn user if task in pipeline does not exist on cluster (#503)

### LINKED ISSUE #503: Warn user if task in pipeline does not exist
We should warn user if the pipeline she is creating contains a task that does not exist on cluster.

lstocchi:
it resolves #503 

Outcome below - a message is also displayed when hovering the task `No task named blabla found on cluster. `
image


lstocchi:
@jeffmaury Now it should work with quoted strings. This needs https://github.com/redhat-developer/intellij-common/pull/128 to work.

jeffmaury (approved):
LGTM even if I hit https://github.com/redhat-developer/intellij-kubernetes/issues/322

PR #509, opened 2021-12-01
fix: prevent resetting whole tree when deleting run (#502)

### LINKED ISSUE #502: Deleting a pipeline run resets the whole tree
Only the PipelineRuns node should be redrawn/repainted.
tekton1

lstocchi:
it resolves #502 
it needs https://github.com/redhat-developer/intellij-common/pull/119 to work


jeffmaury (approved):
LGTM

PR #514, opened 2021-12-14
fix: handle quoted values when detecting unused variables (#513)

### LINKED ISSUE #513: Change behavior of unused variable inspector - underline in yellow
Now unused variable are greyed, we should change this behavior to make it work with current Ij k8s implementation.
Proposal is to underline them in yellow

lstocchi:
it resolves #513


jeffmaury (approved):
LGTM

PR #537, opened 2022-03-11
refactor: rely on k8s plugin for pushing/pulling resources

### LINKED ISSUE #506: Investigate if we can rely on IJ k8s for saving resources
With the new Ij k8s release, we now have the "push" bar above the editor when a yaml resource is edited and needs to be pushed.
We need to verify it covers all use-cases. If so, we can remove our saveListener and then our push dialog.
Example below
image

lstocchi:
it resolves #506


lstocchi:
@jeffmaury removed bump k8s plugin as agreed

jeffmaury (approved):
LGTM

PR #541, opened 2022-03-18
fix: prevent sending telemetry when there is no context

lstocchi:


lstocchi:
@jeffmaury updated as agreed

jeffmaury (approved):
LGTM

PR #551, opened 2022-03-24
chore(skip-release): prepare 1.1.0

lstocchi:
Signed-off-by: Luca Stocchi <lstocchi@redhat.com>


jeffmaury (approved):
I noticed both README.md and plugin.xml previously record changes to the CLI version. Maye we should continue doing so ?

jeffmaury (approved):
LGTM

PR #608, opened 2022-09-20
feat: add support for tekton bundles (#445)

### LINKED ISSUE #445: Tekton bundles
Tekton added a new feature to create an image which has a tekton resource per layer https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#tekton-bundles . 
We should verify who can have access to this feature (only admins can activate it?) and add support to create/push/use a tekton bundle.

lstocchi:
it resolves #445

This patch contains everything to work with tekton bundles. 
The user is able to create a new bundle and push it to a remote registry and then download/import resources from a remote bundle. The gif shows the current workflow.

There could be some optimizations, like download all resources of a bundle with a click, add multiple selection in the new bundle dialog's tree to move resources faster (only single row selection is supported now) but it looks to me a waste of time atm as i foresee very few usage of it. But if you think otherwise I can work on those improvements as well.

There is also an known issue with authentication to a registry. The tkn cli team is fixing it. 
-> https://github.com/redhat-developer/intellij-tekton/issues/609

It needs https://github.com/redhat-developer/intellij-common/pull/162 to work

bundleux

cc @sudhirverma @mohitsuman


jeffmaury:
Latest test remarks:

- multiple elements selection is not available anymore this is very useful when you want to add several items
- so as I need to add element by element;, I used double click but many of them are ignored
- I cannot add more than 10 items where my understanding of the spec is that each layer in the OCI image is limited to 10 elements but an OCI image may have several layers
- While adding the 11th element, I add an error message that pops up but it disappear after a while
- Seems the image name required fullname (tried mybundle or jeffmaury/mybundle) so maybe we could validate it
- tried with docker.io/jeffmaury/mybundle but error message is popped up about missing authentication configuration but disappear as well

lstocchi:
> Latest test remarks:
> 
> * multiple elements selection is not available anymore this is very useful when you want to add several items

I didn't want to spend too much time on something that most likely will not be used. It was an optimization to do when a user would ask for it in my mind. I'll add this.

> * so as I need to add element by element;, I used double click but many of them are ignored

Any example? What are you doing exactly? Which node do you click on?

> * I cannot add more than 10 items where my understanding of the spec is that each layer in the OCI image is limited to 10 elements but an OCI image may have several layers

No, only one resource per layer and max 10 layers. An extract from https://tekton.dev/docs/pipelines/tekton-bundle-contracts/#contract
```
Only Tekton CRDs (eg, Task or Pipeline) may reside in a Tekton Bundle used as a Tekton bundle reference.
Each layer of the image must map 1:1 with a single Tekton resource (eg Task).
No more than 10 individual layers (Pipelines and/or Tasks) maybe placed in a single image.
```

> * While adding the 11th element, I add an error message that pops up but it disappear after a while

Yeah, because there is no other action to remove it otherwise.
You have 10 items, then try to add the 11th. The message pops up. Now when should it be removed? If you remove an item? But it's ugly to keep it fixed imo. You've been informed, if you click it again i show it again what's the issue.

src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCli.java
```diff
@@ -1117,6 +1126,81 @@ private String getFormattedWarningMessage(String reason, String message) {
         return data + "\n";
     }
 
+    public List<String> getResourcesAsYaml(List<Resource> resources) throws IOException {
```
jeffmaury, line 1129: Not sure to understand the rationale for using CompletableFuture as you're synchronizing later to get the result.

src/main/java/com/redhat/devtools/intellij/tektoncd/ui/bundle/BundleUtils.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.regex.Pattern;
+
+public class BundleUtils {
+
+    public static String createCacheKey(String bundle, Resource resource) {
+        return bundle + "-" + resource.type() + "-" + resource.name();
+    }
+
+    public static String cleanImage(String image) {
+        Pattern pattern = Pattern.compile(":.+$");
```
jeffmaury, line 25: Should be defined as a constant for performance reason

src/main/java/com/redhat/devtools/intellij/tektoncd/ui/bundle/CreateBundleDialog.java
```diff
@@ -0,0 +1,305 @@
...
+
+
+    @Override
+    protected void preInit() {
+        bundle = new Bundle();;
+    }
+
+    @Override
+    protected String getTopDescriptionText() {
+        return "Publish a new Tekton Bundle to a registry by passing in a set (max 10) of Tekton objects";
```
jeffmaury, line 96: Maybe replace Tekton objects by Tekton resources

src/main/java/com/redhat/devtools/intellij/tektoncd/ui/bundle/CreateBundleDialog.java
```diff
@@ -0,0 +1,305 @@
...
+        JPanel switchPanel = new JPanel();
+        switchPanel.setLayout(new BoxLayout(switchPanel, BoxLayout.Y_AXIS));
+        switchPanel.add(moveToBundle);
+        switchPanel.add(leaveFromBundle);
+        return switchPanel;
+    }
+
+    private void initActions() {
+        moveToBundleAction = new MoveToBundleAction(bundle,
+                () -> showWarning("The bundle cannot contain more than 10 Tekton objects. " +
```
jeffmaury, line 130: Idem Tekton objects -> Tekton resources. Also layer is probably cryptic for users so I would replace "Please remove some layer before to add new ones" by "Please remove one or several Tekton resources from the bundle before adding new ones"

src/main/java/com/redhat/devtools/intellij/tektoncd/ui/bundle/BundleDialog.java
```diff
@@ -0,0 +1,191 @@
...
+
+    protected void enableLoadingState() {
+        updateLoadingState(new Cursor(Cursor.WAIT_CURSOR));
+    }
+
+    protected void disableLoadingState() {
+        updateLoadingState(Cursor.getDefaultCursor());
+    }
+
+    protected void showTimedErrorMessage(String error) {
```
jeffmaury, line 123: Is it still used ?

jeffmaury (commented):
As a side note, can't we display the bundles as part of the Tekton tree like other resources

src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/TknCli.java
lstocchi: To make calls to cluster async as you may try to download 10 different resources at the same time. Then i need to sync them up as i need all of them to create the bundle... but atleast the download part should be faster than dong it sync, no?

lstocchi:
> As a side note, can't we display the bundles as part of the Tekton tree like other resources

Yes. I didn't add them there because it looks to me a waste of space. Bundles are a bit different compared to other resources. Users should add the nodes (registry/image name) to the tree. Then if someone adds many bundles, everytime the tree is expanded you're going to download everything even if not necessary. is it worth? :man_shrugging:

jeffmaury:
> > As a side note, can't we display the bundles as part of the Tekton tree like other resources
> 
> Yes. I didn't add them there because it looks to me a waste of space. Bundles are a bit different compared to other resources. Users should add the nodes (registry/image name) to the tree. Then if someone adds many bundles, everytime the tree is expanded you're going to download everything even if not necessary. is it worth? 🤷‍♂️

Unless there is a caching mecanism but we can keep it for future

jeffmaury (approved):
LGTM

PR #613, opened 2022-10-12
fix: enable/disable deploy button in create bundle dialog

lstocchi:
it resolves https://github.com/redhat-developer/intellij-tekton/issues/612

### DIFF
--- a/src/main/java/com/redhat/devtools/intellij/tektoncd/ui/bundle/CreateBundleDialog.java
+++ b/src/main/java/com/redhat/devtools/intellij/tektoncd/ui/bundle/CreateBundleDialog.java
@@ -83,6 +83,7 @@ public class CreateBundleDialog extends BundleDialog {
 
     public CreateBundleDialog(@Nullable Project project, Tkn tkn) {
         super(project, "Create and deploy new bundle", "Deploy", tkn);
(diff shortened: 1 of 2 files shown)


jeffmaury (approved):
LGTM but the deploy button is enabled even if the image name is not valid

lstocchi:
> LGTM but the deploy button is enabled even if the image name is not valid

Yeah i made it on purpose so the user can see the error message with the suggestion. Otherwise it is only visible on the tooltip but if you don't pay attention to it and you start typing and you see the button always disabled, you don't get the reason.

jeffmaury:
Yes but seems that on JBT we tried to prevent as much as possible. And I think JetBrains has some kind of recommandations, maybe @adietish may help. Can't you put the error message somewhere else than tooltip ?

lstocchi:
Updated to this
image

adietish:
> Yes but seems that on JBT we tried to prevent as much as possible. And I think JetBrains has some kind of recommandations, maybe @adietish may help. Can't you put the error message somewhere else than tooltip ?


@lstocchi, @jeffmaury: 
> Updated to this image

looks good to me. I'd - like you did - disable "Deploy" when not possible and have an info/warning telling the user about (violated) constraints

PR #615, opened 2022-10-12
chore: bump ij gradle to 1.9.0 (#614)

### LINKED ISSUE #614: bump Gradle IntelliJ Plugin to 1.9.0

lstocchi:
it resolves #614

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,7 +61,7 @@ dependencies {
 }
 
 configurations {
-    compile {
+    implementation {
         exclude group: 'org.slf4j', module: 'slf4j-api'
     }
     integrationTestImplementation.extendsFrom testImplementation
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 
 plugins {
-    id "org.jetbrains.intellij" version "1.7.0"
+    id "org.jetbrains.intellij" version "1.9.0"
     id "idea"
     id "java"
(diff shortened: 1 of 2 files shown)


adietish:
I can reproduce the error that happens in Github actions:
```
> Task :buildSearchableOptions FAILED
2022-10-12 18:07:54,077 [    178]  ERROR - oncurrency.BoundedTaskExecutor - Resource not found: /idea/nullApplicationInfo.xml
java.lang.RuntimeException: Resource not found: /idea/nullApplicationInfo.xml
```
The intellij gradle plugin is the culprit. It works for me when I use version `1.7.0`. It fails when I use `1.8.1` or `1.9.0`.

adietish:
this might be related to https://github.com/JetBrains/gradle-intellij-plugin/issues/1099

lstocchi:
@adietish i don't think it's the same. The issue here also looks different from the one you faced in Ij k8s. Damn MacOs... in my case it's like the system property is not set correctly.
By doing the thing below i overcome the error
```
buildSearchableOptions {
    if (System.getProperty("idea.platform.prefix") == null) {
        systemProperties["idea.platform.prefix"] = ''
    }
}
```

lstocchi:
Closing it as 1.9.0 is not enough for 2022.3. The terminal doesn't work. It needs to be bumped to 1.10

PR #624, opened 2022-11-17
chore: replaced disposable (was: project) & removed TreeHelper.buildModel (#623)

### LINKED ISSUE #623: Remove usage of TreeHelper.buildModel and use correct disposable
related to https://github.com/redhat-developer/intellij-kubernetes/issues/498

* `TreeHelper.buildModel` was needed back then when our least supported version was 2018. We're now requiring at least 2021.1
see here: https://github.com/redhat-developer/intellij-common/blob/main/src/main/java/com/redhat/devtools/intellij/common/tree/StructureTreeModelFactory.java#L39
```java
try {
	Constructor<StructureTreeModel> constructor = StructureTreeModel.class.getConstructor(AbstractTreeStructure.class);
	return constructor.newInstance(structure);
} catch (NoSuchMethodException e) {
	// IC 2019.3+
	Constructor<StructureTreeModel> constructor = StructureTreeModel.class.getConstructor(AbstractTreeStructure.class, Disposable.class);
	return constructor.newInstance(structure, project);
}
```
* `StructureTreeModel` and `AsyncTreeModel` are both provided with the `project` as disposable. The [SDK documentation](https://plugins.jetbrains.com/docs/intellij/disposers.html#choosing-a-disposable-parent) states the `Project` should not be used as disposable because this causes memory leaks:

	> Even though Application and Project implement Disposable, they must NEVER be used as parent disposables in plugin code. Disposables registered using those objects as parents will not be disposed when the plugin is unloaded, leading to memory leaks.

adietish:
fixes #623


src/main/java/com/redhat/devtools/intellij/tektoncd/ui/toolwindow/WindowToolFactory.java
```diff
@@ -56,48 +55,39 @@
...
-            TektonTreeStructure structure = new TektonTreeStructure(project);
-            StructureTreeModel<TektonTreeStructure> model = TreeHelper.buildModel(structure, project);
-            new MutableTektonModelSynchronizer<>(model, structure, structure);
-            Tree tree = new Tree(new AsyncTreeModel(model, project));
-            tree.putClientProperty(Constants.STRUCTURE_PROPERTY, structure);
-            tree.setCellRenderer(new NodeRenderer());
-            ActionManager actionManager = ActionManager.getInstance();
-            ActionGroup group = (ActionGroup)actionManager.getAction("com.redhat.devtools.intellij.tektoncd.tree");
-            PopupHandler.installPopupHandler(tree, group, ActionPlaces.POPUP, actionManager, new TektonTreePopupMenuListener());
+        ContentFactory contentFactory = ContentFactory.SERVICE.getInstance();
```
jeffmaury, line 58: Those kind of changes should be prevented, they make it hard for external reviewers to read the history

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/TreeHelper.java
```diff
@@ -24,67 +22,24 @@
...
-import com.redhat.devtools.intellij.tektoncd.tree.PipelineNode;
-import com.redhat.devtools.intellij.tektoncd.tree.PipelineRunNode;
-import com.redhat.devtools.intellij.tektoncd.tree.ResourceNode;
-import com.redhat.devtools.intellij.tektoncd.tree.TaskNode;
-import com.redhat.devtools.intellij.tektoncd.tree.TaskRunNode;
-import com.redhat.devtools.intellij.tektoncd.tree.TektonRootNode;
-import com.redhat.devtools.intellij.tektoncd.tree.TektonTreeStructure;
-import com.redhat.devtools.intellij.tektoncd.tree.TriggerBindingNode;
-import com.redhat.devtools.intellij.tektoncd.tree.TriggerTemplateNode;
+import com.redhat.devtools.intellij.tektoncd.tree.*;
```
jeffmaury, line 32: Don't use starred imports

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/TreeHelper.java
```diff
@@ -24,67 +22,24 @@
...
 import javax.swing.tree.TreePath;
 import java.io.IOException;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
jeffmaury, line 40: Don't use starred imports

src/main/java/com/redhat/devtools/intellij/tektoncd/utils/TreeHelper.java
```diff
@@ -24,67 +22,24 @@
...
-import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_TASK;
-import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_TASKRUN;
-import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_TASKRUNS;
-import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_TASKS;
-import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_TRIGGERBINDING;
-import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_TRIGGERBINDINGS;
-import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_TRIGGERTEMPLATE;
-import static com.redhat.devtools.intellij.tektoncd.Constants.KIND_TRIGGERTEMPLATES;
-import static com.redhat.devtools.intellij.tektoncd.Constants.NOTIFICATION_ID;
+import static com.redhat.devtools.intellij.tektoncd.Constants.*;
```
jeffmaury, line 42: Don't use starred imports

jeffmaury (changes requested):

src/main/java/com/redhat/devtools/intellij/tektoncd/ui/toolwindow/WindowToolFactory.java
adietish: This change in order is necessary when the disposable to be used is not a parameter any longer but a disposable that's created in the sequence.

jeffmaury: OK I did not catch the reformatting was cause because of the try/catch removal

jeffmaury (approved):
LGTM

Issue #609, opened 2022-09-23
Bundle push not working when authentication is passed through args


lstocchi:
Like explained https://github.com/tektoncd/cli/issues/1718

PR #626, opened 2022-11-21
feat: support authentication when pushing/fetching bundle (#609)

lstocchi:
it resolvs #609


jeffmaury (approved):
LGTM with the following remarks (unrelated to this PR but worth noting):

- if image name contains protocol then an error is spawned when deploying
- version has to be provided we should default it to latest

lstocchi:
> LGTM with the following remarks (unrelated to this PR but worth noting):
> 
> * if image name contains protocol then an error is spawned when deploying
> * version has to be provided we should default it to latest

Opened a new issue to tackle your comment https://github.com/redhat-developer/intellij-tekton/issues/627

PR #632, opened 2022-11-29
fix: change core ns name when working with ocp cluster (#631)

### LINKED ISSUE #631: Unable to open config file on ocp cluster
The plugin looks for configmaps in the `tekton-pipeline` namespace but in ocp this is named `openshift-pipelines`.

lstocchi:
it resolves #631


jeffmaury (approved):
LGTM

PR #640, opened 2022-12-06
chore: add IJ 2022.3 (#639)

### LINKED ISSUE #639: Add IJ 2022.3 verification
IJ 2022.3 is out

lstocchi:
it resolves #639


jeffmaury (changes requested):
I don't see the gradle.bat and grade wrapper jar being updated.
So the proper way to update the Gradle version is the following:

`./gradlew wrapper --distribution-type all --gradle version 7.6`

This is to be executed twice as the second time it will be executed with Gradle 7.6 and thus use the proper script and wrapper JAR

jeffmaury (approved):
LGTM

PR #647, opened 2023-01-02
feat: return v1 snippets if v1 pipelines is installed (#646)

### LINKED ISSUE #646: Use v1 snippets if v1 is enabled on cluster
Now v1beta1 snippets are used by default

lstocchi:
it resolves #646


jeffmaury (approved):
LGTM

PR #661, opened 2024-03-04
chore: updates and fix for 2023.3

sbouchet:


datho7561:
Seems to start up when running with Java 17. I ran into a "wrong class version" error when trying to launch with Java 21, I think this is related to the Gradle version.

sbouchet:
> Seems to start up when running with Java 17. I ran into a "wrong class version" error when trying to launch with Java 21, I think this is related to the Gradle version.

yes, we still use Java 17 as minimal version in all our intellij plugins up to 2023.3.X. Java 21 will be supported with next EAP 2024.X

datho7561 (approved):

PR #669, opened 2024-04-11
chore(skip-release): bump sinceIdea to IU-2022.3

adietish:
Caused following change that requires `sinceIdea` to be bumped to `IU-2022.3`
https://github.com/redhat-developer/intellij-common/pull/219

### DIFF
--- a/.github/workflows/IJ.yml
+++ b/.github/workflows/IJ.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        IJ: [IU-2022.2, IU-2022.3, IU-2023.1, IU-2023.2, , IU-2023.3]
+        IJ: [IU-2022.3, IU-2023.1, IU-2023.2, IU-2023.3, IU-2024.1]
 
     steps:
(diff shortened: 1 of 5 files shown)


adietish:
A test is failing with the bump of base IC version:
`GenericCompletionProviderTest`
```java
   public void testTaskCompletionWithoutAnyInput() {
        assertTrue(getSuggestionsForFile("task1.yaml").isEmpty());
    }
```
`getSuggestionsForFile("task1.yaml")` returns a list with "apiVersion" in it while the test asserts that the list is empty.

adietish:
@sbouchet: I think that we should change the assertion in the test to be valid when "apiVersion" is returned because it's a change in the default behaviour of the platform. Agree?

adietish:
@sbouchet: I introduced a switch in the test that fails starting with `IU-2024.1`:
```java
        if (getIDEAVersion() >= 2024.1f) {
            assertOrderedEquals(suggestions, "apiVersion"); // default completion returns 'apiVersion'
        } else {
            assertTrue(suggestions.isEmpty());
        }
```

sbouchet (approved):
LGTM

